PR #2, opened 2021-09-10
addRank

YagmurSimsekk:


antagomir (changes requested):
Roxygen documentation should be added as well.

R/addRank.R
```diff
@@ -0,0 +1,17 @@
...
+
+setMethod("addRank", signature = c(x = "SummarizedExperiment"),
+          function(x,
+                   method = c("average", "first", "last", "random", "max", "min")){
+          col_data <- colData(x)
+          if (!is.vector(col_data)){
+            col_data <- as.vector(col_data)
+          }
+
+          ranked <- rank(col_data, na.last = TRUE , method)
```
antagomir, line 14: Calling `rank` I think it would be sufficient to provide input arguments as "..."  (`na.last` and `ties.method` can be included that way and not explicitly provided in the function argument by their name). See for instance [getPrevalence.R](https://github.com/microbiome/mia/blob/master/R/getPrevalence.R).

Keep same argument names (ties.method should remain ties.method, not switch to just method).

User needs to have access to changing na.last as well. Giving the "..." as function arguments will do the trick.

R/addRank.R
```diff
@@ -0,0 +1,17 @@
...
+           function(x,
+                    method = c("average", "first", "last", "random", "max", "min"))
+             standardGeneric("addRank"))
+
+setMethod("addRank", signature = c(x = "SummarizedExperiment"),
+          function(x,
+                   method = c("average", "first", "last", "random", "max", "min")){
+          col_data <- colData(x)
+          if (!is.vector(col_data)){
+            col_data <- as.vector(col_data)
```
antagomir, line 11: indentation..? Also run the following ones and make sure there are no errors.

```
document()
run_examples()
check()
BiocCheck::BiocCheck()
```

Then we are already pretty far.

antagomir (changes requested):
Also add a minimal vignette and populate DESCRIPTION file with the necessary information as much as feasible at this point.

YagmurSimsekk:
I changed the addRank.R and tried to do it more precisely and more specific(choosing only integers for now but can be extended to choosing time points since it is aimed to use on time points mostly ) But there is an error that i have been trying to get rid of in 
.is_integer function. That would be so helpful if you have any suggestions

DESCRIPTION
```diff
@@ -1,12 +1,24 @@
 Package: miaTime
 Type: Package
-Title: What the Package Does (Title Case)
+Title: Time Series Analysis
```
antagomir, line 3: -> Microbiome Time Series Analysis

R/addRank.R
```diff
@@ -0,0 +1,48 @@
...
+}
+
+setMethod("addRank", signature(x = "SummarizedExperiment"),
+            function(x,
+                    na.last = "TRUE",
+                    ties.method = c("average", "first", "last", "random", "max",
+                        "min"), ...){
+            col_data <- colData(x)
+            list <- col_data@listData
+            integerlist <- .is_integer(list) #lapply(list, .is_integer)
```
antagomir, line 45: Where does this come from? Are you here trying to test the integer status for multiple different coldata fields?

vignettes/miaTime.Rmd
```diff
@@ -0,0 +1,39 @@
...
+# Installation
+
+To install `miaTime`, install `devtools` first, if it is not installed.
+Afterwards use the `install_github` function from `devtools`.
+
+```{r, eval=FALSE}
+library(devtools)
+devtools::install_github("microbiome/miaTime")
+```
+# Load *miaTime*
```
antagomir, line 29: add space before this line

antagomir (changes requested):

R/addRank.R
YagmurSimsekk: I was trying to do it specific to integers, since it can be ranked correctly in every case. Because if there is something like sampleID for example, the ranking is not necessarily successful because sampleID is nominal. For instance sample1234 is not greater than sample1235 but in ranking it will show like it is. I might be thinking wrong, so in case that's what we want to get anyway , I can do it

antagomir: The idea is to get ranking only for a single selected field that the user would name as an argument. For instance:

`addRank(se, field="time", rank_field="time_rank")`


This would add to `se` a new field `time_rank` and that would contain the rank for the field `time`.

If user wants to do this for different fields, they can do it themselves in they way they need,.

YagmurSimsekk: Thank you for the help! I guess I made it work this time but I am not sure about the function arguments part, especially if there is a need to keep "..."

antagomir: Should work so that if you call 

`addRank(se, field="time", rank_field="time_rank", na.last=TRUE, ties.method="first")`

then if can use all that and the `na.last` and `ties.method` can be passed to `rank` through the "..." mechanism. 

You could ask tvborman for help if not clear otherwise.

YagmurSimsekk:
@antagomir I have updated by adding if statement in case the field that is aimed to be ranked is not in colData. So the field and its ranked field also can be added to colData.

YagmurSimsekk:
@antagomir Also it passed the BiocCheck(), check(), run_examples() without error. Just some warnings and i am trying to clear out those as well

YagmurSimsekk:
I have tried to add citation file but it didn't work out because it is .cff type so it doesn't link automatically. I have tried making cff file but it requires orcid code and doesn't allow me to put my email.

antagomir:
CITATION is normally a plain text file. Check other packages for examples?

R/addRank.R
```diff
@@ -0,0 +1,61 @@
...
+#' @param rank_field_name A character value to name the newly ranked `field`
+#' @param ... Allow new parameters to be defined for this function.
+#'
+#' @return \linkS4class{RangedSummarizedExperiment} object with ranked field
+#' added
+#'
+#' @examples
+#' library(SummarizedExperiment)
+#' data(airway, package="airway")
+#' se <- airway
```
antagomir, line 22: We could prefer our own example data sets but we can leave this for later because mia pkg demo data sets do not have time series currently (we could add?), and using microbiomeDataSets is slow as it has to download data over network

antagomir (changes requested):
devtools::check() gives:

 Undocumented S4 methods:
    generic 'addRank' and siglist 'SummarizedExperiment'
  All user-level objects in a package (including S4 classes and methods)
  should have documentation entries.
  See chapter ‘Writing R documentation files’ in the ‘Writing R
  Extensions’ manual.


-> I think the roxygen is missing for exported functions (note that it is enough to export the user interface functions only, not the internal ones it calls.

antagomir (commented):
Just check also the output of `citation("miaSim")` after `devtools::load_all()` so that you see this and it is OK.

R/addRank.R
```diff
@@ -0,0 +1,61 @@
...
+#'
+#' rank <- addRank(se,field = "SampleName", rank_field_name = "SampleName_rank",
+#'                 na.last = TRUE, ties.method = "first")
+#'
+#' #example 2: new matrix and its rank added
+#' newmatrix <- matrix(1:8, nrow = 8, ncol = 1)
+#'
+#' Added_field <- addRank(se, field = "NewMatrixAdded" ,
+#'             field_matrix = newmatrix, rank_field_name = "newRankedField" ,
+#'             na.last = TRUE , ties.method = "first")
```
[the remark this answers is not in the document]
YagmurSimsekk: Yes that looks better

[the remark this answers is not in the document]
YagmurSimsekk: But should it be added to colData or directly to se?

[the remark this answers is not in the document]
antagomir: `se$newfield ` is in fact adding it to colData. You can check it works that way.

YagmurSimsekk:
> Just check also the output of `citation("miaSim")` after `devtools::load_all()` so that you see this and it is OK.

It seems working for `citation("miaTime")` as well. I was confused just because when i go to "Cite this repository" -> "View citation file" it didn't show me the file unlike in miaSim. It might be because it is still in devel branch but not in main.

antagomir:
You could confirm when this PR is ready to be reviewed again after all changes.

YagmurSimsekk:
@antagomir it is ready

R/addRank.R
```diff
@@ -1,60 +1,51 @@
-#' Rank the sample information in a `SummarizedExperiment` Object
+#' Rank the sample information in a `TreeSummarizedExperiment` Object
```
antagomir, line 1: in fact this could be `SE` because it is more generic than `TreeSE` (TreeSE is an SE), and this function works regardless of tree availability.

R/addRank.R
```diff
@@ -1,60 +1,51 @@
...
-#' The information stored in `colData` in \linkS4class{SummarizedExperiment}
-#' object is ranked by `addRank` function and newly ranked field is added back
-#' to \linkS4class{SummarizedExperiment} object.
+#' The information stored in `colData` in
+#' \linkS4class{TreeSummarizedExperiment} object is ranked by `addRank`
+#' function and newly ranked field is added back to
+#' \linkS4class{TreeSummarizedExperiment} object.
 #'
-#' @param x \linkS4class{SummarizedExperiment} object
+#' @param x \linkS4class{TreeSummarizedExperiment} object containing time series
```
antagomir, line 8: same

R/addRank.R
```diff
@@ -1,60 +1,51 @@
...
 #' @param field A character value indicating the name of the matrix column that
 #' is aimed to be ranked.
-#' @param field_matrix A matrix that is aimed to be ranked. If the aimed matrix
-#' string already exists in colData segment, this optional matrix doesn't need
-#' to be added.
 #' @param rank_field_name A character value to name the newly ranked `field`
 #' @param ... Allow new parameters to be defined for this function.
 #'
-#' @return \linkS4class{RangedSummarizedExperiment} object with ranked field
+#' @return \linkS4class{TreeSummarizedExperiment} object with ranked field
```
antagomir, line 14: same

R/addRank.R
```diff
@@ -1,60 +1,51 @@
...
 #'
 #' @export
 setGeneric("addRank", signature = "x",
-        function(x, field, field_matrix, rank_field_name, ...)
+        function(x, field, rank_field_name, ...)
             standardGeneric("addRank"))
 
-setMethod("addRank", signature(x = "SummarizedExperiment"),
-            function(x, field, field_matrix, rank_field_name, ...){
+setMethod("addRank", signature(x = "TreeSummarizedExperiment"),
```
antagomir, line 43: same

README.Rmd
```diff
@@ -0,0 +1,32 @@
+# Microbiome time series manipulation with miaSim
```
antagomir, line 1: miaSim or miaTime?

README.md
```diff
@@ -1,2 +1,32 @@
-# miaTime
-Time Series Manipulation Package
+# Microbiome time series manipulation with miaSim
```
antagomir, line 1: miaSim or miaTime?

tests/testthat/test-addRank.R
```diff
@@ -1,22 +1,17 @@
...
+    se$newfield <- matrix(1:26, nrow = 26, ncol = 1)
+    se2 <- addRank(se, field = "newfield" ,
+                rank_field_name = "newfield_rank" ,
                 na.last = TRUE , ties.method = "first")
 
-    expect_true(is.matrix(newmatrix))
-    expect_equal(dim(newmatrix), c(8,1))
-    expect_s4_class(Added_field, "RangedSummarizedExperiment")
+    expect_equal(dim(se$newfield), c(26,1))
+    expect_s4_class(se2, "TreeSummarizedExperiment")
```
antagomir, line 15: same

vignettes/miaTime.Rmd
```diff
@@ -1,21 +1,30 @@
...
+```{r, echo=FALSE}
+knitr::opts_chunk$set(cache = FALSE,
+                        fig.width = 9,
+                        message = FALSE,
+                        warning = FALSE)
 ```
 
 `miaTime` implements tools for microbiome time series manipulation based on the
-`SummarizedExperiment` [@SE]  
+`TreeSummarizedExperiment` [@TSE]
```
antagomir, line 27: same

antagomir (commented):
We will still need to change the example data. Working on it.

YagmurSimsekk:
All corrected

DESCRIPTION
```diff
@@ -1,12 +1,34 @@
...
+              comment = c(ORCID = "0000-0001-5537-637X")),
+       person(given = "Yagmur", family = "Simsek", role = c("aut", "cre"),
+              email = "yagmur.simsek.98@gmail.com"))
+Description: 
+    Time series analysis of microbiome data sets is evaluated by the functions
+    in `miaTime` package.
+biocViews: Microbiome, Software, Sequencing, Coverage
+License: Artistic-2.0 | file LICENSE
+Depends:
+    SummarizedExperiment
```
FelixErnst, line 17: Add the R version. I am not sure whether it is a Bioc requirment, but it good form. (Only the major not the minor version)

DESCRIPTION
```diff
@@ -1,12 +1,34 @@
...
+Imports:
+    methods,
+    TreeSummarizedExperiment
+Suggests: 
+    rmarkdown,
+    knitr,
+    mia,
+    devtools,
+    BiocStyle,
+    testthat (>= 3.0.0)
```
FelixErnst, line 27: There are no specific features in testthat 3.0, which you are using/probably should be using. I would remove the version info here

DESCRIPTION
```diff
@@ -1,12 +1,34 @@
...
 Encoding: UTF-8
-LazyData: true
-RoxygenNote: 7.1.1
+URL: https://github.com/microbiome/miaTime
+LazyData: false
+Roxygen: list(markdown = TRUE)
+RoxygenNote: 7.1.2
+VignetteBuilder: knitr
+Config/testthat/edition: 3
```
FelixErnst, line 34: I would remove this not standard config info

FelixErnst (commented):
Hi @YagmurSimsekk 

The `addRank` name is not quite clear in the context of `mia`. taxonomic rank usually live in the `rowData`, but you are working on the `colData` with other rank information.

Is rank used in the meaning of sequence and order? If yes `addOrderSequence` mnight be better name.

Felix

R/addRank.R
```diff
@@ -0,0 +1,52 @@
...
+
+setMethod("addRank", signature(x = "TreeSummarizedExperiment"),
+            function(x, field, rank_field_name, ...){
+            col_data <- colData(x)
+            mdat <- do.call(cbind,lapply(col_data, as.vector))
+            data <- mdat[, field]
+            ranked <- rank(data, ...)
+            col_data[, rank_field_name] <- ranked
+            colData(x) <- col_data
+            return(x)
```
FelixErnst, line 51: Shorter and faster:

```
col_data <- colData(x)[,field] # this will error if field is not present. still the length of field needs to be tested in an input test first
colData(x)[, rank_field_name] <- rank(data, ...)
```

antagomir:
> The `addRank` name is not quite clear in the context of `mia`. taxonomic rank usually live in the `rowData`, but you are working on the `colData` with other rank information. Is rank used in the meaning of sequence and order? If yes `addOrderSequence` mnight be better name.

Yes it is used in that meaning and I agree there is potential confusion with `rowData`.

Let us change the name to `addOrderSequence`.

DESCRIPTION
antagomir: We also seem to have the first 3 of these in `mia` and `miaViz` DESCRIPTION files. Those to be removed as well?

R/addRank.R
YagmurSimsekk: thank you!

antagomir: Can you "resolve" this particular conversation (and others) as they are resolved?

FelixErnst:
Or maybe `addTimeRank` or something like this, if you want to keep the rank part in the function name

antagomir:
> Or maybe `addTimeRank` or something like this, if you want to keep the rank part in the function name

I think `addOrderSequence` is better since we may potentially use this also for other ordered features than just time, and rank is indeed easily confused with `rowData`.

In fact, this is the first step. As a second step we will need a solution where rank (of time points) is given per subject (or another named colData category). This will allow us to pick the first, second, or Nth time point across all subjects. A simple example would be to subset the data with the baseline (first) time point per subject.

FelixErnst:
This is an interesting concept.

If you want to got deep into that, I think you should consider a new class derived from `DataFrame`, which holds only `Date` types. With that setup, you could put the DataFrame in the colData (DataFrame supports nesting) and be very flixble with how many time series you have and what type of data you need (time differences, ranks, etc can calculated on the fly from `Date`).

For Inspiration you should take a look at the `DotBracketDataFrame` class in the `Structstrings` package.

antagomir:
Thanks for the suggestion, I think we should explore this. 

However, this may take a longer time to investigate fully. Meanwhile, a simple method to extract the baseline for each subject would be very useful.

A limitation can be that time is just one aspect that can be used for sorting.

antagomir:
@YagmurSimsekk can you look at this - you can use the example data set already in *miaTime*

```
library(miaTime)
data(hitchip1006)
colData(hitchip1006)
colData(hitchip1006)$time
```

The example linked by Felix is worth exploring and examples welcome, but equally important is to get forward.

At the minimum, prepare an example that converts a given (time) field into a `Date` object and then uses readily available functions for those objects to return rank index and time differences. Then we could provide a wrapper that adds these into the colData after user specifies time field/s.

YagmurSimsekk:
@antagomir I am not sure if the numeric values in time field in hitchip1006 indicated the minutes/seconds or anything else, but I did it assuming that these are seconds. 
`HitchipDate <- TimePointsToDate(hitchip1006, "time", origin = "1970-01-01")`
`> HitchipDate@colData`
DataFrame with 1151 rows and 13 columns
It includes extra Date(POSIXct), rank and time difference fields.

antagomir:
These are days (can you update to that?), also highlighting the challenges. Sometimes time is given in unitless forms. Would be good to support that as well, perhaps a purely numeric field. Will check soon more closely.

R/addOrderSequence.R
```diff
@@ -0,0 +1,52 @@
...
+#' se1 <- addOrderSequence(se,field = "SampleType",
+#'             rank_field_name = "SampleType_rank",
+#'             na.last = TRUE, ties.method = "first")
+#'
+#' #example 2: new matrix and its rank added
+#'
+#' se$newfield <- matrix(1:26, nrow = 26, ncol = 1)
+#' se2 <- addOrderSequence(se, field = "newfield" ,
+#'                 rank_field_name = "newfield_rank" , na.last = TRUE ,
+#'                 ties.method = "first")
```
antagomir, line 39: Could we also have example with a numeric or date field for `addOrderSequence`? Either from GlobalPatterns or from hitchip1006.

R/addOrderSequence.R
```diff
@@ -0,0 +1,52 @@
...
+#'
+#' @export
+setGeneric("addOrderSequence", signature = "x",
+        function(x, field, rank_field_name, ...)
+            standardGeneric("addOrderSequence"))
+
+setMethod("addOrderSequence", "TreeSummarizedExperiment",
```
antagomir, line 46: would it be enough to have `SE`?

README.Rmd
antagomir, line 1: miaTime?

antagomir (commented):
Looks good for getting started with the discussion on details.

antagomir:
The `TimePointsToDate` is now adding multiple fields. Can we just have one function per added field? And if we can utilize existing time series tools from other packages that's fine, then just making examples on those in examples field and vignette is sufficient. One row to convert to dates (if necessary..?), one row to add rank, one row to add time differences.

R/TimePointsToDate.R
```diff
@@ -0,0 +1,36 @@
...
+    ifelse(seconds[i] > 0.6, add_minutes <- seconds[i] %/% 0.6, add_minutes)
+    minutes <- add_minutes + minutes
+    seconds <- seconds - add_minutes*0.6
+
+    seconds <- round(seconds, 2)*10
+    ct <- as.POSIXct(60 * minutes, origin, tz = "UTC") + seconds
+    colData(data)$Date <- ct
+    colData(data)$rank <- rank(colData(data)[,t])
+    z <- diff.POSIXt(ct)
+    colData(data)$timeDifference<- c(NA,z)
```
antagomir, line 34: I think we should have calculation of differences available as its own separate function. 

We also should not make assumptions about the time unit but the user should be able to choose. This is easier if the time difference function is separated in its own function. Then user can just use the information from Date field, and use already existing tools from other packages to add the differences. We can provide examples for that, and if it seems that we can provide a wrapper function that makes this easier, then it might be worth implementing. But let us now hardcode critical choices on the time unit, like this, within the functions.

R/TimePointsToDate.R
```diff
@@ -0,0 +1,36 @@
+#'Numerically provided time points are converted to a POSIXct object in the
+#'format "%Y-%m-%d %H:%M:%OS".
+#'
+#'@param data : dataset with time points in it
+#'@param t : character time points field in the given dataset
+#'@param origin : character indicating the start of the time points
+#'(default: \code{origin = "1970-01-01"})
```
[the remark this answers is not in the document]
YagmurSimsekk: The POSIXct object has to have an origin and if we want to consider hours/minutes/seconds we should use `POSIXct ` rather than `Date`. I couldn't find datasets with dates in mia or in curatedMetagenomiceData but GrienesenTSData has it. I can get the origin from there, but it is already given as `Date` object, so it doesn't help to turn numeric value to `Date` as we wanted.

R/TimePointsToDate.R
```diff
@@ -0,0 +1,36 @@
...
+#'data(hitchip1006)
+#'force(hitchip1006)
+#'HitchipDate <- TimePointsToDate(hitchip1006, "time", origin = "1970-01-01")
+#'
+#'@return a \linkS4class{TreeSummarizedExperiment} object containing added
+#'date object, rank of the time points and time difference between the time
+#'points in seconds.
+#'
+#'@export
+TimePointsToDate<- function(data, t, origin = "1970-01-01"){
```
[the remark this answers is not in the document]
YagmurSimsekk: We can make them choose the units of the given time points or like you suggested we can extract the only time from POSIXct object using  `strftime(times, format="%H:%M:%S")`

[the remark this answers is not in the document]
antagomir: Ok great. 

My choice would be to recognize that there are at least 3 feasible ways to encode time information: 1) Date 2) Time as in POSICXct object (`02:06:49`) and 3) plain numeric.

The plain numeric is often used in many external time series analytics tools, and cases 1-2 can converted into it.

My suggestion would be to encourage the use of cases 1-2 where possible, to include metadata; and then have examples (or even wrappers) showing how to convert these into a plain numeric format. Then the numeric format would be the starting point for most statistical/computational analyses.

The advantage of this is that Date/POSIX allow various conversions and utilities for time data that are not available for the numeric format, and the concept of time units is also more clearly encoded in these. But the numeric format is often the de facto standard for final analyses.

My suggestion is therefore to pick for instance HITChip data, add an example that shows how to convert the current numeric time field (days) into the POSIX format as one colData field, and an example how to convert from that POSIX format back to a numeric field.

YagmurSimsekk:
@antagomir Sorry, I had deadlines in the weekdays, so it took a while. 
I ended up creating a separate function to convert `"00:07:09"` object to numeric value based on selected time unit; hours, minutes or seconds. After conversion, the difference is recalculated according to the selected time unit. Also, the baseline can be changed, but by default it is set to the first time point. Finally, I decided to convert them to list object and show how to apply them to SE object.

The only downside is that `"00:07:09"` is no longer a POSIX object, because POSIX objects need a date appended. (1970-01-01) In this case it is just a character class object.

Also, I'm not sure if `addOrderSequence `matches other functions. Because it applies the rank field directly to the SE. Also, it's just one base function used (rank), so I'm not sure if it's better to keep it separate there or use it inside other functions.

antagomir:
[This thread](https://stackoverflow.com/questions/22659947/r-how-to-handle-times-without-dates) has suggestions for handling times without dates. It proposes **chron** and **lubridate** packages. After some reading, my impression is that [lubridate](https://lubridate.tidyverse.org/) might be the way to go. For some more exmaples, see [this](https://r4ds.had.co.nz/dates-and-times.html), [this](https://www.neonscience.org/resources/learning-hub/tutorials/dc-convert-date-time-posix-r) and one example on a Bioconductor package [CellBench](https://bioconductor.org/packages/release/bioc/vignettes/CellBench/inst/doc/Timing.html).

Could lubridate provide a solution?

Yes, `addOrderSequence` is just one liner with `rank`. This could be left out but in the next step we want to use these tools to do similar operations _per subject_. Perhaps we can come back to this function then.

YagmurSimsekk:
**lubridate** converts to "Period" class object, it looks quite different (`"7H 58M 0S"`) . Also, it only works on character value, for example 
```
z <- Sys.time()
> class(z)
[1] "POSIXct" "POSIXt" 
> hms(z)
[1] NA 
```

with characters, 
```
x <- c("09:10:01", "09:10:02", "09:10:03")
y <- hms(x)
[1] "9H 10M 1S" "9H 10M 2S" "9H 10M 3S"
> class(y)
[1] "Period"
attr(,"package")
[1] "lubridate"
```
chron package does exactly the same and it converts to "times" class object. 

```
> tt <- times(c("12:13:14", "15:46:17"))
> ss <- trunc(tt, times("00:01:00"))
> ss
[1] 12:13:00 15:46:00
> class(ss)
[1] "times"
```

Addition to that, there is also **data.table** that creates "ITime" class object. But at least it converts from POSIXct object. 

```
> z <- Sys.time()
> as.ITime(z)
[1] "10:38:44"
> class(z)
[1] "POSIXct" "POSIXt"
```

Just to have an idea, there is also POSIXlt object that can extract each component of the time. 
```
> z <- Sys.time()
> d <- as.POSIXlt(z)
> d$mday
[1] 14
> d$min
[1] 43
```

FelixErnst:
`lubridate` is a really useful package, but I would only use it, when the default `Date` type cannot be used.

`chron` is a package dating back quite some time, actually it reimplements some ideas from S. I don't want to say it is bad, because it is old, but the ideas of code and data structures in Bioc might not be as compatible.

The main requirement is, thatr time can be stored as columns in a `DataFrame` without breaking or handling being very complicated. If that is the case, IMO it is safe to assume, that SE objects can be enriched with time data of a certain type.

antagomir:
The main issue is that `Date` objects require the availability of date information, and very commonly this is not available in experimental studies (or their recorded metadata) because the date is not always considered an important part of the experimental design. The use of `Date` objects can difficult or misleading in such cases, where we do not have a date, and hence would not want to assign date, to time fields.

The lubridate package might provide one solution, not sure if that is the optimal one. 

Encoding time as a numeric value is one option that may be often needed in practice (for various time series analysis algorithms etc). Regardless of the solution ,we will therefore need a way to convert the time field information to numerics, in order to support many external algorithms.

We can consider supporting many different time formats, like `Date`, **lubridate**, `numeric` etc. but the implementation can become messy. Users can choose to use whatever they wish in `colData` but for the miaverse framework it might be best to choose just 1-2 solutions that are actively supported.

If `Date` objects have obvious problems, we need to either come up with a solution to circumvent the incorporate of date information in `Date` objects or come up with an alternative solution. The alternative could be one of the other available time classes (lubridate or other). It would be beneficial to support a commonly used time format. But if there is no good general solution, one option is just to assume that users operate with date/time information outside of _miaverse_, and time would be presented in a simple numeric format in _miaverse_.

YagmurSimsekk:
So, the solution to this problem can be using lubridate package to convert time information given in character to "period" class object. Also, there is `as.POSIXlt()` that can break the time information into its units and then these units can be stored based on choice. ( date can be excluded in case it is not available). 

And there should be a function that converts time fields given in different formats (may be limited to the most commonly used time formats) to numeric and saves them in numeric format.

antagomir:
In general, it might be a good idea to support that users will encode time information with time-specific classes that have clear units. At the same time, many existing algorithms require time as a numeric vector. We would need a way to easily convert between numeric time (with a given unit) and Date/lubridate, both ways. 

If this is easily achieved, we should support using the Date/lubridate in metadata (including examples on how to get into that from numeric data), and make sure that conversion to numeric is easy, when needed. The `as.POSIXlt()` seems useful. I am hesitating whether we should additionally support using numeric as an optional way of representing time data.

I would consider supporting time in both numeric and  in Date/lubridate formats in downstream analysis functions.

Can we start by making a simple example on how to convert numeric time data into colData with `as.POSIXlt()` and then how to use that information for instance to add rank as a new field to colData? As another example, changing the baseline time and calculating differences from baseline time point.

In a second step, we could proceed to splitting time series per subject, for instance to pick baseline time point per subject, but this can be another issue.

YagmurSimsekk:
Regarding the last commit; I found out that `as.POSIXlt` and lubridate package will basically do the same since both break down  time to its units. So I used lubridate and now the numeric time field can be "Period" class and optionally, if the date information is known, that can be added as well. Also, numeric value can be converted to character since it is one of the commonly used formats for time. "Period" and character to numeric conversion are shown in another function: `timeNumeric`

It can also be applied to any time vector, no longer just limited to the data given in the SE container.

R/timeNumeric.R
```diff
@@ -0,0 +1,49 @@
+#' Time field provided in character as "%H:%M:%S" or in Period class ("H M S")
+#' can be converted to numeric values depending on the units.
+#'
+#' @param time_field : vector containing the time points
+#' @param units : The unit of time points selected in `time_field`.
+#' It can be hours, minutes or seconds.(default: \code{unit = "mins"})
+#'
+#' @examples
+#' data(hitchip1006)
+#' hitchipTime <- timeToDate(time_field = colData(hitchip1006)[,"time"],
```
antagomir, line 10: hitchip time is in days, fix here and in other places

R/timeNumeric.R
```diff
@@ -0,0 +1,49 @@
+#' Time field provided in character as "%H:%M:%S" or in Period class ("H M S")
```
antagomir, line 1: Is this really `character`, rather than `Date`? Visual appearance is the same but the Date objects have time handling tools, characters do not.

antagomir (commented):
See the minor comments. I will now test the examples in more detail.

R/timeNumeric.R
antagomir: I can now see that this format does not support days. Is this something we cannot overcome easily?

R/timeDifferences.R
```diff
@@ -0,0 +1,35 @@
...
+#' #\linkS4class{SummarizedExperiment} object as new fields.
+#'
+#' colData(hitchip1006)$time_diff <- diffHitchip$Consecutive
+#' colData(hitchip1006)$base_diff <- diffHitchip$Baseline
+#'
+#' @return a list object containing consecutive time differences and
+#' the distance from selected time point to each time point.
+#'
+#' @export
+timeDifferences <- function(time_field, baseline = 1){
```
antagomir, line 25: The baseline shift is covered already by lubridate (`x - x[[1]]`). This should be sufficient, with an example in OMA on how to add this field to colData.

The time difference is covered already by lubridate `diff(x)`. This should be sufficient, with an example in OMA on how to add this field to colData.

Conclusion: this function is not needed as such. 

However, we would like to do this per subject. I suggest to remove this function from the current PR and merge the other things when they are ready. Then in a separate PR, let us create a new function that can split samples per subject, add these informations per subject, and then collect the data back to the original colData. This will complicate the present PR too much, therefore let's remove this function from this PR.

R/addOrderSequence.R
```diff
@@ -0,0 +1,57 @@
+#' Rank the sample information in a `SummarizedExperiment` Object
```
antagomir, line 1: As far as I can see, an equivalent of this function can be implemented with just one line:

`colData(hitchip1006)$rank <- rank(colData(hitchip1006)$time)`

This would be more transparent, flexible, and does not require maintenance. 

Hence I suggest to remove `addOrderSequence` from this PR.

However, we may like to do a subject-specific version of this for a new PR. Let us rethink that more properly in a next PR after merging the current one. Remove this function from the current PR.

antagomir (changes requested):
Thanks @YagmurSimsekk - I went now through these in more detail. 

Let us remove some functions from this PR as proposed in the comments so that we can get this merged. Let us avoid reinventing the wheel whereever we can. Let us finalize this PR with the Period class (I think that was now the choice?), then we can proceed to implementing subject specific functions and other things.

YagmurSimsekk:
> This is an interesting concept.
> 
> If you want to got deep into that, I think you should consider a new class derived from `DataFrame`, which holds only `Date` types. With that setup, you could put the DataFrame in the colData (DataFrame supports nesting) and be very flexble with how many time series you have and what type of data you derive from thatg (time differences, ranks, etc can calculated on the fly from `Date`).
> 
> For Inspiration you could take a look at the `DotBracketDataFrame` class in the `Structstrings` package.

@FelixErnst Hi, i have looked at the package and could see what you meant, but when DataFrame nested, its class changes. Here in `Structstrings` package, it is  `DotBracketDataFrame` and there are similar nested DataFrame classes such as `SimpleDFrameList` or `CompressedSplitDFrameList` but none of them can be implemented in colData field in (Tree)SE object since it accepts only DFrame. Is there a way to overcome this issue?

FelixErnst:
You can add a `CompressedSplitDFrameList` as column of a DataFrame. 

```r
> df <- DataFrame(a = c(1,2,3))
> df$b <- list(df,df,df)
> df$c <- DataFrameList(df,df,df)
> df
DataFrame with 3 rows and 3 columns
          a      b                       c
  <numeric> <list>         <DataFrameList>
1         1  1,2,3 1:1,2,3,2:1,2,3,3:1,2,3
2         2  1,2,3 1:1,2,3,2:1,2,3,3:1,2,3
3         3  1,2,3 1:1,2,3,2:1,2,3,3:1,2,3
> class(df$c)
[1] "SimpleDFrameList"
attr(,"package")
[1] "IRanges"
> df$d <- SplitDataFrameList(df,df,df)
> class(df$d)
[1] "CompressedSplitDFrameList"
attr(,"package")
[1] "IRanges"
> se <- SummarizedExperiment(assays = list(counts = matrix(c(1,2,3), ncol=3)))
> se
class: SummarizedExperiment 
dim: 1 3 
metadata(0):
assays(1): counts
rownames: NULL
rowData names(0):
colnames: NULL
colData names(0):
> colData(se) <- df
> colData(se)
DataFrame with 3 rows and 4 columns
          a      b                       c
  <numeric> <list>         <DataFrameList>
1         1  1,2,3 1:1,2,3,2:1,2,3,3:1,2,3
2         2  1,2,3 1:1,2,3,2:1,2,3,3:1,2,3
3         3  1,2,3 1:1,2,3,2:1,2,3,3:1,2,3
                                                                                                d
                                                                             <SplitDataFrameList>
1 1:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3,2:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3,3:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3
2 1:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3,2:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3,3:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3
3 1:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3,2:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3,3:1,2,3:1:1,2,3,2:1,2,3,3:1,2,3
```

So this example shows that you can nest DataFrames as much as you like.

In case of `Structstrings` I chose to limit certain things by create a special classes such as `DotBracketDataFrame` and the related `CompressedSplitDFrameList` classes.

PR #3, opened 2021-09-19
documenting the hitchip data

antagomir:


antagomir:
Can be merged if you agree. This contains data that we can use to test the new methods that will be migrated in here.

PR #8, opened 2021-11-25
pkgdown added

YagmurSimsekk:


.github/workflows/pkgdown.yaml
```diff
@@ -0,0 +1,33 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/master/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+on:
+  push:
+    branches: [main, master]
```
FelixErnst, line 5: Either or. not both this will get you into trouble

PR #9, opened 2021-11-27
timePerSubject

YagmurSimsekk:
Still figuring out where to put the list object in SE... Also thought that, this can be used on more than just subject, could be more generalized like using any information given, even integers like age etc. can be converted to factor, then can be used here. But, I assume, then it wouldn't be in the concept of miaTime.

Also, the other thing I realized is that functions get really limited in this use. For example, "order" can be chosen but descending order  `order(... , decreasing = TRUE ) ` cannot be applied. Should we try to overcome this?


R/timePerSubject.R
```diff
@@ -0,0 +1,57 @@
...
+timePerSubject <- function(se, time_field, subject_field, operation, new_field){
+
+    colData(se)$time_field <- colData(se)[,time_field]
+
+    colData(se)$subject_field <- colData(se)[, subject_field]
+
+    colData(se) <- colData(se) %>%
+        as.data.frame() %>%
+        group_by(col = subject_field) %>%
+        mutate(new_field = match.fun(operation)(time_field)) %>%
```
YagmurSimsekk, line 42: @antagomir here i tried to change to `mutate(new_field = match.fun(operation)(time_field, ...))` and also added ` ...` to arguments

R/timePerSubject.R
```diff
@@ -0,0 +1,57 @@
...
+#' @param new_field : character class object refers to the name of a
+#' newly added field
+#'
+#' @examples
+#' library(miaTime)
+#' data(hitchip1006)
+#' se <- hitchip1006
+#'
+#' hitchipTime <- timePerSubject(se, time_field = "time",
+#'     subject_field = "subject", operation ="rev", new_field = "reversed")
```
YagmurSimsekk, line 20: also added `decreasing = TRUE ` to end and didnt' work.  `error: x unused argument (decreasing = TRUE)`

YagmurSimsekk (commented):
This is how i tried to pass other arguments by `...` but didn't work

antagomir:
@YagmurSimsekk can we close this one?

PR #14, opened 2021-12-07
Vignette

antagomir:


antagomir:
I added various examples on the planned operations that do not need their own implementations.

R/getTimeDivergence.R
```diff
@@ -0,0 +1,54 @@
...
+
+    timedivergence_n <- matrix(NA, ncol = 1, nrow = length(n))
+    timedifference_n <- matrix(NA, ncol = 1, nrow = length(n))
+
+    for (i in location){
+        for (j in seq_len(length(location))){
+            timedivergence_n[i] <- x[j]
+            timedifference_n[i] <- time_diff[j]
+        }
+    }
```
YagmurSimsekk, line 51: Actually, the function doesn't work properly. After the for loop, it returns only the last value for every row that I wanted to assign a value (except NAs) but I am working on it and I wanted to share anyway to see if I am on the right track. After I solve this problem, I will add the samples and the functions from mia.

vignettes/miaTime.Rmd
```diff
@@ -95,20 +117,153 @@ Timediff <- c(NA, Timediff)
...
+
+# Check output
+as.data.frame(colData(se)[, c("subject", "time", "shifted")]) %>%
+    dplyr::filter(subject == "831")
+```
+
+
+## Estimating divergence and convergence
+
+[@gutMicrobiota] is available for further details.
```
YagmurSimsekk, line 190: I couldn't access to full article, so I couldn't find the relation and left it like this for now.

YagmurSimsekk (commented):
Not complete yet and open to any suggestions/feedbacks

antagomir:
> 

Are there specific questions?

Issue #13, opened 2021-12-07
Calculate beta diversity between samples at the interval of n time steps


antagomir:
Time series analyses often measure dissimilarity (beta distance) between consecutive time steps, within n=1, 2, 3, or more steps

In this case, beta diversity needs to be calculated between all the given sample, and the other sample that precedes it n steps in time. We are here referring to time steps because the interval is not always uniform.

*Suggestion* let's implement a function that calculates beta diversity in time series between samples that have time difference of n time steps. Also calculate information of the difference in actual time between those same time steps. 

Return the result as a data.frame; default field names could be `timedivergence_n` and `timedifference_n` where n is replaced with the time step interval (1,2,3...). The resulting dataframe has one dissimilarity per sample; and the first n samples will have NAs for obvious reason that they do not have preceding samples n steps before them in the time series. 

In addition, show in the examples how that can be added to  colData.  

The dissimilarity calculation must have access to all beta diversity measures available in mia. Start with `mia::calculateDistance`, later add `calculateJSD` and `calculateUnifrac` as options if possible.

Suitable function name (to start with at least): `getTimeDivergence`.

PR #16, opened 2021-12-10
getTimeDivergence

YagmurSimsekk:


R/getTimeDivergence.R
```diff
@@ -0,0 +1,54 @@
...
+    n <- 1:length(colData(se)[, time_field])
+
+    if (!(time_interval %in% n)){
+        stop("Time interval cannot be bigger than the length of time series")
+    }
+
+    vector <- colData(se)[, time_field]
+    location <- seq(starting_position, length(vector), time_interval)
+
+    x <- (vector)[location]
```
antagomir, line 39: What does `x` stand for here? Time divergence should be calculated as beta diversity but as far as I can see, this is not beta diversity?

R/getTimeDivergence.R
```diff
@@ -0,0 +1,78 @@
...
+    #related samples are chosen from assay
+    x <- as.matrix(assay(tse)[,rownames(sample_df)[which(is.na(mat[, time_field]) == FALSE)]])
+
+    if(!transposed){
+      x <- t(x)
+    }
+
+    timedivergence_n <- as.matrix(vegan::vegdist(x), "bray")
+
+    #timedivergence_n <-do.call(stats::dist, list(x))
```
YagmurSimsekk, line 69: I am not sure which function does the job, but I assume ` vegan::vegdist` it is.

R/getTimeDivergence.R
```diff
@@ -0,0 +1,78 @@
...
+    #timedivergence_n <- as.data.frame(as.matrix(time_divergence_n))
+
+    time_diff <- diff(extracted_time)
+
+    timedifference_n <- c(NA, time_diff)
+
+    return(list(timedivergence_n = timedivergence_n,
+                timedifference_n = timedifference_n))
```
YagmurSimsekk, line 77: I couldn't put them in data.frame since the dimensions are different. It might be that I am still missing out something.

antagomir: Which dimensions are different? Not sure if I follow, good to check first.

R/getTimeDivergence.R
antagomir: The functions in mia package are (as I also wrote in the inital issue #13 )

https://github.com/microbiome/mia/blob/master/R/calculateDistance.R

-> This is using vegdist but this might be more handy to use with (TreeSE). Therefore I proposed to use this function.

After calculateDistance works, then these two would be good to add:
- https://github.com/microbiome/mia/blob/master/R/calculateJSD.R
- https://github.com/microbiome/mia/blob/master/R/calculateUniFrac.R

R/getTimeDivergence.R
YagmurSimsekk: #16 you said that I can put them in a data.frame and columns would be `timedivergence_n` and `timedifference_n `  but i didn't get there.

antagomir: Right, I see.

Between time point 1 and 2 there will be one time difference (time step = 1, or time unit (days?)= xxx). 

Between time point 1 and 2 there will be one beta diversity difference (or divergence) ie. one similarity score between the two samples from time points 1-2.

Hence, for every time step, we have one time difference and one divergence. 

I am not sure which dimensions are different?

antagomir:
These are now handled in vignette examples in PR #14

R/getTimeDivergence.R
```diff
@@ -58,21 +60,24 @@ getTimeDivergence <- function(se, sample_field, sample_id, time_field, time_inte
...
-    x <- as.matrix(assay(tse)[,rownames(sample_df)[which(is.na(mat[, time_field]) == FALSE)]])
+    x <- as.matrix(assay(se)[,rownames(sample_df)[which(is.na(mat[, time_field]) == FALSE)]])
 
     if(!transposed){
       x <- t(x)
     }
 
-    timedivergence_n <- as.matrix(vegan::vegdist(x), "bray")
+    #timedivergence_n <- calculateDistance(x)
+    timedivergence_n <- calculateJSD(x)
```
YagmurSimsekk, line 71: Should `calculateJDS` and `calculateDistance ` be used here in the same function? If yes, then an optional parameter can be placed to choose which one to use and there could be a default function as well. The other function `calculateUniFrac` works differently and cannot be used in the same way as others. I am trying to find different ways to make it applicable in this function.

R/getTimeDivergence.R
```diff
@@ -58,21 +60,24 @@ getTimeDivergence <- function(se, sample_field, sample_id, time_field, time_inte
...
+                                     FUN.VALUE = 1)
 
-    timedifference_n <- c(NA, time_diff)
 
-    return(list(timedivergence_n = timedivergence_n,
-                timedifference_n = timedifference_n))
+    mat.m$sample_id <- c(sample_id)
+    mat.m <- mat.m[,c("sample_id", "sp1", "sp2", "dist", "timedifference_n")]
+
+    return(mat.m)
```
YagmurSimsekk, line 82: Return is a matrix containing the samples used in the beta diversity and the time difference between them. I am still not sure how to implement it back to colData since each row has 2 samples that are being considered for beta diversity calculation. The only way I can think is to have separate colData field from the original one.

R/getTimeDivergence.R
antagomir: Yes, ideally we could choose use all of these common beta diversity methods also with time series.

But already a proper implementation with `calculateDistance` would be a valuable addition. Can start with that if adding others is too complex in the first stage. Another option is that we have separate functions for these different beta diversities, with uniform naming style.

R/getTimeDivergence.R
antagomir: There is no need to return sample profiles to colData. The colData should have 1) beta diversity i.e. difference between sample T and sample T-t; 2) time difference between time steps for sample T-t. Both of these are univariate.

antagomir: Here T meaning the latter and t meaning the former sample. By default, consecutive time steps but we also allow longer intervals (differences between two time steps or more, optionally).

antagomir: Sample profiles we already have in the abundance table (assay), not sure why they would be needed in colData.

YagmurSimsekk: Then the comment about colData in #13 is meant different from what I understood. 

Time difference between T-t is there. For 1) beta diversity between sample T and sample t is there too, but that's not what you meant, I guess. Okay, let me give an example. For instance, here in this example, Baboon-1 has 6 different time steps. In the case of time interval = 2, only the samples at t = 2, t = 4, t= 6 used for beta calculation. So the beta calculation is between t2-t4, t2-t6 and t4-t6 and the time difference between these also calculated. And sp1 and sp2 fields in return matrix show the samples used in the beta diversity.

antagomir: With time interval = 2, we need the beta diversities for t2/t4, t3/t5, t4/t6 i.e. all pairs that are separated by 2 time steps. The pair t2-t6 in your example above is separated by 4 time steps.

YagmurSimsekk:
Even though it works right now, especially with Grienesen data, I am still not so satisfied with the efficiency of code. So I will try with the new data set as soon as I create SE for that, and I will come back to function and fix if needed.

R/getTimeDivergence.R
```diff
@@ -0,0 +1,98 @@
...
+        list_t <- lapply(seq_len(length(time)), function(i){ as.matrix(assay(se)[,names(samplename[[i]])])})
+
+        if(!transposed){
+          list_t <- lapply(seq_len(length(list_t)), function(i){
+            t(list_t[[i]])
+          })
+        }
+
+        timedivergence_n <- vapply(seq_len(length(list_t)),
+                                   FUN = function(i) {calculateDistance(list_t[[i]])},
```
antagomir, line 79: The latest version of *mia* says

Warning message:
'calculateDistance' is deprecated. 
Instead, use directly the function that is specified by 'FUN' argument. 

The FUN argument refers to `stats::dist` 

-> Let us use that one.

Then, with the time interval t, should this be just simple as:

```
t <- 1 
mat <- matrix(1:60, nrow = 5) 
timedivergence_n <- sapply((t+1):nrow(mat), function (i) {stats::dist(mat[c(i, i-t), ])})
```

antagomir: (or something like that for `vapply`)

R/getTimeDivergence.R
```diff
@@ -0,0 +1,98 @@
...
+                                   FUN.VALUE = 1)
+
+
+        timedifference_n <- vapply(seq_len(length(timedivergence_n)),
+                                   FUN = function(i) {diff(time[[i]])},
+                                   FUN.VALUE = 1)
```
antagomir, line 85: Shouldn't this be just like 

timedifference_n <- sapply((t+1):nrow(mat), function (i) {diff(df[c(i, i-t), time_field]))})

R/getTimeDivergence.R
```diff
@@ -0,0 +1,98 @@
...
+        time <- mat[, time_field][location]
+        mat_t <- as.matrix(assay(se)[,names(samplename)])
+        if(!transposed){
+          mat_t <- t(mat_t)
+        }
+        timedivergence_n <- calculateDistance(mat_t)
+        timedifference_n <- diff(time)
+    }
+
+    return(cbind(timedivergence_n,timedifference_n))
```
antagomir, line 97: It would be handy to have additionally a function `estimateTimeDivergence` that adds the output from `getTimeDivergence` to input `SE` object's `colData` directly. In the same way than in `mia::estimateDiversity`.

antagomir (changes requested):

R/getTimeDivergence.R
```diff
@@ -0,0 +1,98 @@
...
+#'
+#' The dissimilarity (beta distance) is calculated in steps of n between the
+#' samples of subject. "n" can be 1 or greater than 1. The time difference
+#' between the n points is also calculated.
+#'
+#' @param se A
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
+#' object or
+#' \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#' @param sample_field column vector containing the all individuals in
```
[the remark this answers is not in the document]
YagmurSimsekk: `sample_field` refers to all individuals, but I am now deleting the `sample_id` field, so it would calculate for all individuals. Maybe the `sample_id` can be an optional parameter in case the dataset is long.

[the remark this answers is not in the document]
antagomir: I would rather give the user subset the data before using this function if they need to do this for only some limited individuals. The typical use case would be to this for all individuals. 

However, this probably should be implemented so that the function 1) splits the data by individual (or another grouping); 2) performs the operation per individual (or another group); 3) combines the individual results again into a single object.
This would be a robust way of implementing this. Also compare to [lines 141-155 in vignette](https://github.com/microbiome/miaTime/blob/master/vignettes/miaTime.Rmd). The main difference is that the vignette example operates only on `colData` whereas in the current case we need both `colData` and `assay` data -> Can we split the whole SE object by individual (or another group! so that this is not specific to "individuals" but more general to any grouping factor); then calculate the operation per each SE object; then collect the splitted SE objects back to a single SE object?

R/getTimeDivergence.R
YagmurSimsekk: I have been trying to implement the method which splits SE by chosen field, applies the operation, and puts it back. But there is an error that I couldn't fix for a while. Here I used `time` as the field to split SE object, but it can be anything as you suggested. 

```
data(hitchip1006)
se <- hitchip1006
field <- "time"

spl <- split(colnames(se), colData(se)[, field])
  
se_list <- lapply(seq_len(length(spl)), function (i) {SummarizedExperiment(assays = as.matrix(assay(se)[,spl[[i]]]),
                                                                             colData = colData(se)[spl[[i]],])})
   
mat <- sapply(seq_len(length(se_list)), function(i){t(assay(se_list[[i]]))})
   
t <- 1

sapply(seq_len(length(mat)), function(n){ 
        sapply((t+1):nrow(mat[[n]]), function (i) {stats::dist(mat[[n]][c(i, i-t), ])})})
```
error is in `sapply` part, I tried with for loop, and it gives the same error. `Error in mat[[n]][c(i, i - t), ] : subscript out of bounds`
 ` i ` stays in 2L and `s` stays in 6L, and it is because mat[[6]] has only one sample. Then I tried something like this 

`sapply(seq_len(length(mat)), function(n){ 
      if( nrow(mat[[n]]) > 1 ){
          sapply((t+1):nrow(mat[[i]]), function (i) {stats::dist(mat[[i]][c(i, i-t), ])}
          )
      } else if ( nrow(mat[[n]]) == 1 ) { 
        next
        }
  }
)`

That also didn't work. I am out of solutions currently. Anyhow, I can do the same operation without splitting the SE but instead making data frame from the chosen field.

antagomir: Ok how about this one:

antagomir:
```
library(miaTime)
data(hitchip1006)
se <- hitchip1006
field <- "subject"


# Define the time step of interest
step <- 1
myassay <- "counts"

# Define the names for the newly added fields
# for the time difference and community dissimilarity
new_field <- "time_divergence"
new_field2 <- "time_difference"

# Dissimilarity will be calculated between rows, as in stats::dist
check_pairwise_dist <- function (x, distfun, step, myassay, new_field) {

    mat <- t(assay(x, myassay))

    # Add new field to coldata
    colData(x)[, new_field] <- rep(NA, nrow(mat))
    
    if (nrow(mat) > step) {
         inds <- (step+1):nrow(mat)
         d <- sapply(inds, function (i) {distfun(mat[c(i, i-step), ])})
	 colData(x)[inds, new_field] <- d
    }
    return(x)
}

# Split SE into a list, by subject
spl <- split(colnames(se), colData(se)[, field])
se_list <- lapply(seq_len(length(spl)), function (i) {se[, spl[[i]]]})

# Manipulate each subobject
se_list <- lapply(seq_along(se_list),
    function(i){ check_pairwise_dist(se_list[[i]], distfun=stats::dist, step, myassay, new_field) })

# Merge the objects back into a single SE
se <- do.call("cbind", se_list)
```

antagomir: It is a bit slow, but acceptable for a first version. Time difference to be added, and material to be polished.

YagmurSimsekk: Thanks! I'll complete what is missing and look for a faster implementation.

YagmurSimsekk: I have just realized that it gives the same error. Because again, `x` is only one subject, so the distance cannot be applied in this way.

YagmurSimsekk:
```
inds <- (time_interval+1):length(se_list)
      d <- sapply(inds, function (i) { s <- t(assay(se_list[[i]]))
                                       ss <- t(assay(selist[[i-time_interval]]))
                                       dd <- rbind(s, ss)
                                       distfun(dd))})
```
But I found a different way which does the job. I will add everything soon.

YagmurSimsekk:
There is another unclear point on my part. It is that after splitting SE by subject, some list object have more than one sample.
 `spl <- split(colnames(se), colData(se)[, field])
> spl[["217"]]
[1] "Sample-217"  "Sample-1119" "Sample-1120"`

So I changed the function to calculate the diversity only between the list items that have one sample and for others `timedivergence_n` and `timedifference_n `assigned as NAs. (for implementation in colData). Could there be a better solution?

antagomir:
But if the split is done per subject, and the idea is to calculate community divergence between time points _within a subject_ then I think the only place where you can do that is in those lists that have more than one sample, i.e. those subjects that have multiple time points. This function should not calculate community divergences between the different elements of the splitted list because each element corresponds to a different subject (or another grouping field). And we are not calculating this between subjects, only within subjects.

YagmurSimsekk:
It makes a lot more sense. I'll change the function this way.

YagmurSimsekk:
`gitTimeDivergence` works well. I could use it with different data sets. 
`
spl <- split(colnames(se), colData(se)[, field])
spl[["217"]]
[1] "Sample-217" "Sample-1119" "Sample-1120"

colData(hitchipTime)[217,]
                      time      sample        time_divergence     time_difference
                    <numeric> <character>       <numeric>       <numeric>
Sample-217        0.0  Sample-217              NA              NA
Sample-1119       2.3 Sample-1119         3291.68             2.3
Sample-1120       8.3 Sample-1120         3956.36             6.0
`

However it throws error when going through `devtools::check()` . 
Error in (function (classes, fdef, mtable)  : 
  unable to find an inherited method for function ‘bindCOLS’ for signature ‘"TreeSummarizedExperiment"

I tried to fix it but couldn't fix any other method to combine TSE objects that are in list together.

antagomir:
Can you use `SEtools::mergeSEs(se_list)` instead of cbindand also add SEtools to Imports in DESCRIPTION
seems to solve the issue

README.md
```diff
@@ -18,15 +21,32 @@ devtools::install_github("microbiome/miaTime")
...
+  
+   checking DESCRIPTION meta-information ...
+  
+✔  checking DESCRIPTION meta-information
+## 
+  
+─  checking for LF line-endings in source and make files and shell scripts
+## 
+  
+─  checking for empty or unneeded directories
```
antagomir, line 41: What are these doing in README.md?

R/getTimeDivergence.R
```diff
@@ -0,0 +1,105 @@
...
+#'                                    time_field = "time")
+#'
+#'@export
+getTimeDivergence <- function(se, field, time_field, time_interval){
+
+# Define the names for the newly added fields
+# for the time difference and community dissimilarity
+new_field <- "time_divergence"
+new_field2 <- "time_difference"
+
```
antagomir, line 47: these should be removed, or given as function arguments?

R/getTimeDivergence.R
```diff
@@ -0,0 +1,105 @@
...
+# Dissimilarity will be calculated between rows, as in stats::dist
+check_pairwise_dist <- function (x,
+                                 distfun,
+                                 time_interval,
+                                 new_field,
+                                 new_field2,
+                                 time_field){
```
antagomir, line 54: This is internal function and should be moved outside of `getTimeDivergence`? See *mia* pkg how internal functons are handled.

R/getTimeDivergence.R
```diff
@@ -0,0 +1,105 @@
...
+
+    mat <- t(assay(x))
+
+    time <- colData(x)[, time_field]
+
+    # Add new field to coldata
+    colData(x)[, new_field] <- rep(NA, nrow(mat))
+    colData(x)[, new_field2] <- rep(NA, nrow(mat))
+
```
antagomir, line 63: remove double empty lines

R/getTimeDivergence.R
```diff
@@ -0,0 +1,105 @@
...
+#' @importFrom SummarizedExperiment colData
+#' @importFrom SummarizedExperiment colData<-
+#'
+#' @examples
+#' library(miaTime)
+#' data(hitchip1006)
+#'
+#' hitchipTime <- getTimeDivergence(hitchip1006, field = "subject",
+#'                                    time_interval = 1,
+#'                                    time_field = "time")
```
antagomir, line 38: Pick a subset of data to speed up example. This example is rather slow due to large number of samples.

R/getTimeDivergence.R
```diff
@@ -0,0 +1,105 @@
+#' Beta diversity calculation within individuals in the interval of n time steps
+#'
+#' The dissimilarity (beta diversity) is calculated in time steps of n, between
+#' the samples of subject(individuals).Time interval can be 1 or greater than 1.
```
antagomir, line 4: should have space before "("

R/getTimeDivergence.R
```diff
@@ -0,0 +1,105 @@
...
+
+# Dissimilarity will be calculated between rows, as in stats::dist
+check_pairwise_dist <- function (x,
+                                 distfun,
+                                 time_interval,
+                                 new_field,
+                                 new_field2,
+                                 time_field){
+
+    mat <- t(assay(x))
```
antagomir, line 56: assay(x) does not specify which abundance values ie. counts, relabundance, clr.. other..? should be passed as function argument. Also check if there are other similar cases in the code.

R/getTimeDivergence.R
```diff
@@ -0,0 +1,105 @@
...
+    return(x)
+}
+
+# Split SE into a list, by subject
+spl <- split(colnames(se), colData(se)[, field])
+
+# Manipulate each subobject
+se_list <- lapply(seq_along(spl),
+                  function(i){check_pairwise_dist(x = se[, spl[[i]]],
+                                                   distfun=stats::dist,
```
antagomir, line 93: I suggest to replace `stats::dist` with `vegan::vegdist` everywhere.

antagomir (changes requested):
Getting there.

Can you run `tools::resaveRdaFiles("./")` to resave compressed data.

Unit tests should be added.

R/getTimeDivergence.R
```diff
@@ -33,73 +33,92 @@
...
+getTimeDivergence <- function(se, field, time_field, time_interval, new_field = "time_divergence",
+                              new_field2 = "time_difference", abund_values = "counts" ){
+
 # Split SE into a list, by subject
 spl <- split(colnames(se), colData(se)[, field])
 
+spl_more <- spl[lapply(spl,length)>1]
+
+spl_one <- spl[lapply(spl,length) == 1]
+
```
YagmurSimsekk, line 94: I tried to make the function faster and came up with this. Here are the individuals that have more than one sample separated and instead of all of them, only they are applied to `check_pairwise_dist` .  The ones that have only one sample goes through the `se[, spl_one[[i]]]` later on. Eventually these lists are put together and combined as SE object. Hitchip1006 would run in 180 seconds for the entire dataset, but the old version was running it for 250 seconds. Actually, it didn't save time as much as I hoped, so we can go back to the previous version if you think that it was more robust/efficient.

antagomir: Ok, just let's try to get something now that does the job and works as expected (unit tests). Optimization can be done later.

Anyway take a subset of hitchip for examples, examples should always be very fast.

antagomir:
Checks fail, fix?

YagmurSimsekk:
Checks pass. Ready to merge?

antagomir:
Thanks, I will test soon.

Issue #22, opened 2022-02-07
vignette example: returning baseline time point


antagomir:
Add the following data subsetting example in miaTime vignette, under a section subtitle "Subset to baseline samples". HITChip data can be used an example.

Example 1:
- identify samples with time point 0
- subset the data to those samples

Example 2:
- identify samples with the smallest time point value within each subject (using the earlier examples adding the rank of each time point within subject)
- subset the data to those samples (picking the rank=1 for each subject)

Example 3:
- same as Example 2 but with/out NA time points (na.rm is an option in the rank method I think); when we apply this including NA time points, we should also get one sample for those subjects that are lacking time point information

PR #31, opened 2022-02-26
subset baseline samples

YagmurSimsekk:


YagmurSimsekk:
Third example in #22 is not possible with HITChip data since it doesn't have any NA time points for any subject. Should I use a different data set for this issue?

antagomir:
Write there 

```
# Add some NAs in the time point data to demonstrate functionality with missing data
nainds <- sample(seq_len(ncol(hitchip1006)), 50)
colData(hitchip1006)$time[nainds] <- NA
```

vignettes/miaTime.Rmd
```diff
@@ -215,7 +230,7 @@ BiocManager::install("mia")
 ```
 
 
-```{r, eval=FALSE}
+```{r eval=FALSE, include=FALSE}
```
antagomir, line 233: could we have eval=TRUE here and elsewhere? would be better for constant testing, if possible

YagmurSimsekk: That was because I thought it would give error due to mia dependency. It is removed now, ready to be tested.

antagomir:
Seems to go through nicely!

PR #39, opened 2022-05-12
new minimal microbiome data

microsud:
Added the 16 species minimal microbiome data
README fixed link to miaverse project update
added Cairo and ggplot2 to suggests to avoid note/warnings


antagomir (commented):
Awesome! 

The only thing I was still thinking is the data set name, is there anything a bit more descriptive than "mdbMM16"? Some options might be: "Shetty2022" (we have often named data sets like this); or "minimalgut" or so. Aliases are also possible to have. Just a thought.

microsud:
I agree. Will change it to `minimalgut`

Issue #64, opened 2022-12-26
getStepwiseDivergence extension to altExp/reducedDims


antagomir:
getStepwiseDivergence picks data  from assays; in some use cases it would be useful to apply this to reducedDims (trajectory plots in ordination). 

At the same go, good to consider the use of altExps.

Should be considered more thoughtfully before implementation.

Afterwards, update examples on ordination trajectory plots in miaTime vignettes.

antagomir:
Add same support also for other functions where relevant?

antagomir:
Essentially, in `.check_pairwise_dist` function, recognize if `assay_name` is available in assays or reducedDim. 

Then pick the data from the respective one to the `mat` matrix. Other changes should not be needed in the function itself. Once `mat` has been defined either from assay or altExp, everything should work as expected.


If the same `assay_name` is used in both `assay` and `reducedDim`, then pick it from the `assay` and generate warning that announces this problem.

Update documentation, explaining that `assay_name` can be used also to refer to reducedDim entries. And that assay_name has priority if these two have overlapping names.

Add sufficient unit tests accordingly for the different combinations as this can be a nasty one to catch otherwise-

antagomir:
After done, consider the need to add similar support for `getBaselineDivergence`. Open an issue if relevant.

ChouaibB:
One question about altExp, shall it be an option to choose the assay from as well (along with the option of reducedDim)? or shall altExp be part of the implementation of choosing from assay or reducedDim?
I have a draft of an implementation coming-up with the assay and reducedDim options only so far.

antagomir:
I think (but I am not sure) that reducedDim applies to the full TreeSE object, and is not altExp-specific. Correct if I am mistaken here.

When altExp support is added, the assay should be according to the specified altExp. If altExp is not specified, then the standard assay from TreeSE is used. You could check from other functions in mia how they deal with the altexp argument (for instance runNMDS but perhaps good to check a couple of them in case there are alternative solutions; could even compare to other external packages like scater to make sure the treatments are unified).

PR #70, opened 2022-12-29
Issue 64

ChouaibB:
This is related to #64 
All implementation so far has been only within `.check_pairwise_dist`


R/getStepwiseDivergence.R
```diff
@@ -22,14 +22,18 @@
...
+#' It could also be used for `reducedDims` entries. Yet, the ones present in 
+#' `assays` are the priority.
 #' @param FUN a \code{function} for dissimilarity calculation. The function must
 #'   expect the input matrix as its first argument. With rows as samples 
 #'   and columns as features. By default, \code{FUN} is
 #'   \code{vegan::vegdist}.
 #' @param method a method that is used to calculate the distance. Method is
 #'   passed to the function that is specified by \code{FUN}. By default,
 #'   \code{method} is \code{"bray"}.
+#' @param altexp String or integer scalar specifying an alternative experiment
```
antagomir, line 35: an -> the

R/getStepwiseDivergence.R
```diff
@@ -222,3 +233,36 @@ setMethod("getTimeDivergence",
     return(x)
 
 }
+
+############################ HELPER FUNCTION(S) ################################
+
+.get_mat_from_se <- function (x, assay_name, altexp) {
```
antagomir, line 239: Is this code borrowed from some other package?

R/getStepwiseDivergence.R
```diff
@@ -22,14 +22,18 @@
 #' samples used to calculate beta diversity
 #' (default: \code{name_timedifference = "time_difference"})
 #' @param assay_name character indicating which assay values are used in
-#' the dissimilarity estimation (default: \code{assay_name = "counts"})
+#' the dissimilarity estimation (default: \code{assay_name = "counts"}).
+#' It could also be used for `reducedDims` entries. Yet, the ones present in 
+#' `assays` are the priority.
```
antagomir, line 27: Modify the text as: "This argument can be used to specify data set available through assay or reducedDim slot to be used in divergence calculations. In the case of name conflict (same name in both assay and reducedDim), the assay will have priority."

antagomir (approved):
Looks good! Just make sure that the tests cover all combinations of altexps, assays, and reducedDims.

R/getStepwiseDivergence.R
ChouaibB: No, I made it myself for the task.

R/getStepwiseDivergence.R
```diff
@@ -222,3 +233,36 @@ setMethod("getTimeDivergence",
...
+    mat <- NULL
+    if (assay_name %in% assayNames(x)) {
+        mat <- t(assay(x, assay_name))
+    }
+    # x has to be an instance of SingleCellExperiment to have a reducedDim slot
+    if (is(x, "SingleCellExperiment") && assay_name %in% reducedDimNames(x)) {
+        if (!is.null(mat))
+            # if present in both assay and reducedDim
+            warning("'assay_name' was present in assay and reducedDim. ",
+                    "'assay_name' is picked from assay.",
```
antagomir, line 257: I would check immediately at the start of the script to avoid any unnecessary further steps.

If available in both -> use assay, provide warning
If available in just one -> pick accordingly

There could be helper function for picking each so that it is nicely splitted in components.

R/getStepwiseDivergence.R
```diff
@@ -222,3 +233,36 @@ setMethod("getTimeDivergence",
...
+    if(!.is_non_empty_string(assay_name)){
+        stop("'assay_name' must be a non-empty single character value.",
+             call. = FALSE)
+    }
+    if (!is.null(altexp)) {
+        x <- altExp(x, altexp)
+    }
+    mat <- NULL
+    if (assay_name %in% assayNames(x)) {
+        mat <- t(assay(x, assay_name))
```
antagomir, line 250: just wondering is it really necessary to transpose (and convert the matrix into unexpected way)

R/getStepwiseDivergence.R
```diff
@@ -222,3 +233,36 @@ setMethod("getTimeDivergence",
...
+    }
+    # x has to be an instance of SingleCellExperiment to have a reducedDim slot
+    if (is(x, "SingleCellExperiment") && assay_name %in% reducedDimNames(x)) {
```
antagomir, line 253: is(x, "SingleCellExperiment") also good to check as the first thing, nothing else will make sense if this is FALSE

R/getStepwiseDivergence.R
```diff
@@ -222,3 +233,36 @@ setMethod("getTimeDivergence",
...
+                    "'assay_name' is picked from assay.",
+                    call. = FALSE)
+        else
+            # if not present in assay but in reducedDim
+            mat <- SingleCellExperiment::reducedDim(x, assay_name)
+    }
+    # not present otherwise
+    if (is.null(mat))
+        stop("'assay_name' is not present.",
+             call. = FALSE)
```
antagomir, line 266: Isn't this something that could be checked already on the very first lines? If assay_name is not in assayNames and not in reducedDimNames then one can immediately stop with this warning? 

Doing this at the end via investigation of matrix contents is unnecessarily indirect.

R/getStepwiseDivergence.R
antagomir, line 239: Should this be enable also for case where altexp=NULL (?)

antagomir: Ok looks neat. Just compare to the earlier implemented solutions in mia, and make sure you are not reinventing new solutions (unless they are clearly better and thus justified).

antagomir (changes requested):

R/getStepwiseDivergence.R
ChouaibB: I would ofc mention it was adapted from.... as a comment or put it at utils.R header with the other imports from other packages.

antagomir: Which package it was adapted from ?

ChouaibB: I meant in case if I would have copy pasted the code from another place or package, I would mention as a comment that it was so , or adapted from and mention the source. If it was imported from another package, I will do the import at utils.R for instance of the package.
In this case, I have not copy pasted the code from anywhere, I built based on the steps required for the task.

antagomir: Ok. Just that we always have to take care of open license compatibility issues when reusing code.

R/getStepwiseDivergence.R
ChouaibB: if this is FALSE then the check of assay_name on reducedDimNames won't evaluate.
In case, the altExp is a SummarizedExperiment it won't have a reducedDim slot then will pass it. Will check only for those who have a reducedDim slot e.g.  SingleCellExperiment, TreeSummarizedExperiment.

R/getStepwiseDivergence.R
ChouaibB: intention was not to break the existing flow (if I understood right the task), since afterwards rows are supposed to be samples. And assay call to the resulting x or altExp(x) would produce features as rows.

R/getStepwiseDivergence.R
antagomir: Ah right.

R/getStepwiseDivergence.R
antagomir: In general, in TreeSE assays, rows are features and columns are samples.

R/getStepwiseDivergence.R
ChouaibB: I guess that's what's going on atm (or perhaps I didn't understand the comment right).
first check name in assay (and pick it up already if so)
second check if in reducedDim (if mat is picked already, then throw warning otherwise pick it up if positive)
Third if nothing was picked then stop.

R/getStepwiseDivergence.R
antagomir: Resolve conversations once done.

R/getStepwiseDivergence.R
ChouaibB: Yes, but when mat is returned from the function then line 212 assumes mat having samples on rows.

R/getStepwiseDivergence.R
antagomir: I mean, could it be checked at the very first lines of this function already, that an assay (or reducedDim) with such name exists in the named altExp? And then proceed to pick the object that had that name.

-> Instead of running the whole function with various if/else/pick statements, and only in the end noticing that the result is still empty.

R/getStepwiseDivergence.R
antagomir: as long as there are no unnecessary transposes it is fine

antagomir: in principle, it would be most clear if the function that should pick SE assay, would return it in the same way than it is contained in the actual assay.

And then if vegan functions need transpose, it could be transposed during that function call.

R/getStepwiseDivergence.R
ChouaibB: Hi,
I have just noticed that [mia imports](https://github.com/microbiome/mia/blob/master/R/utils.R#L5) a function from scater [.get_mat_from_sce](https://github.com/Alanocallaghan/scater/blob/master/R/utils_reddim.R#L3).
Shall this be rather used adding to the `getStepwiseDivergence` two more arguments; `dimred` and `n_dimred`?

antagomir: Sounds good!

R/getStepwiseDivergence.R
```diff
@@ -76,11 +83,21 @@ getStepwiseDivergence <- function(x,
                             name_divergence = "time_divergence",
                             name_timedifference = "time_difference",
                             assay_name = "counts",
-			    FUN = vegan::vegdist,
-			    method="bray", ...){
+			                FUN = vegan::vegdist,
+            			    method="bray",
+            			    altexp = NULL,
+			                dimred = NULL,
+			                n_dimred = NULL,
```
antagomir, line 90: align the rows

antagomir (approved):
Looks good!

PR #80, opened 2023-07-27
Rename agglomerateByRank

Daenarys8:
See https://github.com/microbiome/mia/issues/392


antagomir (approved):

antagomir:
Is there a deprecation message, so that the old function name will still continue working with a warning? (if not, possible to check examples from other functions)

antagomir:
@Daenarys8

Issue #89, opened 2024-08-19
Auto-correlation


TuomasBorman:
Consider adding method for calculating auto-correlation?

antagomir:
Best strategy might be to provide a general wrapper for some time series pkg that can calculate many different time series measures, not only autocorrelation but other measures, too.

PR #90, opened 2024-08-26
add ts measures

0xMuluh:
ping: #89 
Added wrapper for acf, pacf, arima and box.test


R/getTsMeasure.R
```diff
@@ -0,0 +1,80 @@
...
+        }
+        
+        ############################ Input check end ###########################
+        
+        ############################ Measure Calculation ######################
+        # Initialize a list to store results
+        results <- list()
+        
+        # Loop through the specified measures and calculate each
+        for (measure in measures) {
```
TuomasBorman: Use lapply instead

R/getTsMeasure.R
```diff
@@ -0,0 +1,80 @@
...
+            if (measure == "acf") {
+                results$acf <- acf(ts_data, plot = FALSE)$acf
```
TuomasBorman: The calculation part could be moved to internal function

TuomasBorman: Also try to use switch() whereever possible (or other that kind of structure). It makes the code cleaner when the calculation is done in one step

Consider this kind of structure:
```
exportedFUN <- function(){

res <- lapply(methods, function(method){
.internal_fun(method)
}
names(res) <- measures
return(res)
}


.internal_fun <- function(measure){
# Get correct function to call
FUN <- switch()
# Calculate
res <- FUN()
# Get the result
if( measure %in% c("acf", "pacf") ){
res <- res$acf
}
return(res)
}
```

R/getTsMeasure.R
0xMuluh: ACK

R/getTsMeasure.R
0xMuluh: ACK

0xMuluh:
@TuomasBorman @antagomir

R/getTsMeasure.R
```diff
@@ -0,0 +1,86 @@
...
+#' @export
+setGeneric("getTsMeasure", signature = "x",
+    function(x, ...) 
+       standardGeneric("getTsMeasure")
+)
+
+#' @rdname getTsMeasure
+#' @export
+setMethod("getTsMeasure", signature = c(x = "TreeSummarizedExperiment"),
+    function(x, time.field = "Time.hr", measures = "acf", ...) {
```
TuomasBorman, line 33: As time.field can be anything, I don't believe there is benefits by setting the default value. So time.field could not have any value bu default. That might be more clear

R/getTsMeasure.R
```diff
@@ -0,0 +1,86 @@
...
+#' from a specified column in the `colData` of a `TreeSE` object.
+#' 
+#' @param x A
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
+#' object.
+#' @param time.field \code{Character scalar}. Specifies the name of the
+#' time series field in `colData`. (Default: \code{"Time.hr"})
+#' @param measures \code{Character vector}. Specifies the time series measures 
+#' to compute. Supported measures include "acf", "pacf","Box.test", and "arima".
+#' (Default: \code{ ""acf""})
```
TuomasBorman, line 12: ""acf"" --> "acf"

R/getTsMeasure.R
```diff
@@ -0,0 +1,86 @@
...
+                data.", call. = FALSE)
+    }
+    
+    # Get the correct ts measure to call
+    FUN <- switch(measure,
+                  "acf" = function() acf(ts_data, plot = FALSE),
+                  "pacf" = function() pacf(ts_data, plot = FALSE),
+                  "Box.test" = function() Box.test(ts_data, type = "Ljung-Box"),
+                  "arima" = function() arima(ts_data),
+                  stop("Unsupported measure: ", measure, call. = FALSE))
```
TuomasBorman, line 75: Pass additional arguments with "..."

R/getTsMeasure.R
TuomasBorman, line 12: Documentation is lacking info. For instance, arima model has lots of parameters that user wants to modify. However, now it is very hard to know what are those parameter names

TuomasBorman (changes requested):
Unfortunately, I’ve only had time to review this function thoroughly now, and it seems that it's not working correctly. The issue is that it calculates measures/models for time point values(?), which isn't meaningful. Instead, it should calculate the values based on abundances. Additionally, there needs to be a way to group values based on patient.

It seems that the functions you are utilizing expect the input format to have rows representing time points and columns representing patients.
To fix this:

1. Modify the abundance table so that rows represent time points and columns represent patients.
2. Check if missing time point values are handled by the functions by default.
3. Utilize the functions. Pass additional parameters using `...`
5. Improve the documentation, as it doesn’t currently describe which functions are used. Each of these methods has a unique set of parameters that need to be clarified.

antagomir:
Thanks. I suggest that we hold dealing with this PR until other critical issues have been dealt with. This is extra functionality which is not urgent.

antagomir:
I will close this now. Analysis and visualization functionality should be kept separate. We can reopen this when the need for ts measure functionality is clarified. 

We could discuss more in #89 before deciding *how* to implement this.

antagomir:
Or alternatively, we can limit the number of measures to few that are easy to handle well, if this PR was close to be ready.

PR #91, opened 2024-09-16
refactor getBaselineDivergence/getStepwiseDivergence

### LINKED ISSUE #58: Optimize getBaselineDivergence
This function is too slow for larger data sets and should be optimized,

0xMuluh:
Dependent on https://github.com/microbiome/mia/pull/635


0xMuluh:
Given that the calculation has changed, should we still support the baseline_sample parameter as either a vector or a SummarizedExperiment object?

@TuomasBorman 
@antagomir

TuomasBorman:
If I remember correctly addDivergence supports
- Variable from colData
- Numeric vector denoting reference sample
- Character vector denoting reference samples for each sample
- Character scalar denoting a reference sample

The same should/could be supported in miaTime (perhaps use the same function)? However, to make things simpler, we should prioritize the colData variable in the documentation, otherwise it might become too complex.

I think the only difference between addDivergence and addBaselineDivergence is that the latter adds the time difference?

addStepwiseDivergence is little bit more complex since we should specify the previous step/sample for each sample. But after that, the function is same as addDivergence (apart from adding time difference).

0xMuluh:
@antagomir

R/deprecate.R
```diff
@@ -0,0 +1,52 @@
...
+#' @rdname deprecate
+#' @export
+setGeneric("getTimeDivergence", signature = c("x"), function(x, ... )
+    standardGeneric("getTimeDivergence"))
+
+#' @rdname deprecate
+#' @export
+setMethod("getTimeDivergence", signature = c(x = "ANY"), function(x, ...){
+    .Deprecated(msg = paste0("'getTimeDivergence' is deprecated. ",
+                             "Use 'addStepwiseDivergence' instead."))
```
TuomasBorman: No need to use paste0() in these deprecation messages

R/getBaselineDivergence.R
```diff
@@ -6,9 +6,20 @@
...
 #'
-#' @inheritParams getStepwiseDivergence
-#' @param baseline_sample \code{Character vector}. Specifies the baseline sample(s) to be used. If the
-#'   "group" argument is given, this must be a named vector; one element per group. (Default: \code{NULL})
+#' @inheritParams addStepwiseDivergence
+#' @param FUN \code{Function} for dissimilarity calculation. The function must
+#' expect the input matrix as its first argument. With rows as samples and 
+#' columns as features. (Default: \code{vegan::vegdist})
+#' @param altexp \code{Character scalar} or \code{integer scalar}. Specifies the 
+#' alternative experiment containing the input data. (Default: \code{NULL})
```
TuomasBorman: This might be good to move as hidden parameter as this is not core functionality

R/getBaselineDivergence.R
```diff
@@ -6,9 +6,20 @@
 #' The method operates on `SummarizedExperiment` objects, and the results
 #' are stored in `colData`.
 #'
-#' @inheritParams getStepwiseDivergence
-#' @param baseline_sample \code{Character vector}. Specifies the baseline sample(s) to be used. If the
-#'   "group" argument is given, this must be a named vector; one element per group. (Default: \code{NULL})
+#' @inheritParams addStepwiseDivergence
+#' @param FUN \code{Function} for dissimilarity calculation. The function must
```
TuomasBorman: This should be the same name as in mia::addDivergence

R/getBaselineDivergence.R
```diff
@@ -6,9 +6,20 @@
...
+#' expect the input matrix as its first argument. With rows as samples and 
+#' columns as features. (Default: \code{vegan::vegdist})
+#' @param altexp \code{Character scalar} or \code{integer scalar}. Specifies the 
+#' alternative experiment containing the input data. (Default: \code{NULL})
+#' @param dimred \code{Character scalar} or \code{integer scalar}. indicates the 
+#' reduced dimension result in `reducedDims` to use in the estimation. 
+#' (Default: \code{NULL})
+#' @param n_dimred \code{Integer vector}. Specifies the dimensions to use if
```
TuomasBorman: Should be ndimred or other than is used in mia

R/getBaselineDivergence.R
```diff
@@ -18,16 +29,58 @@
...
+#' (per group). Alternatively,
+#' the user can provide the baseline vector, or a list of baseline vectors per
+#' group (named list per group).
 #'
-#' @importFrom mia mergeSEs
+#' @examples
+#' library(miaTime)
+#'
+#' data(hitchip1006)
+#' tse <- mia::transformAssay(hitchip1006, method = "relabundance")
```
TuomasBorman: Load mia and use it without :::

R/getBaselineDivergence.R
```diff
@@ -18,16 +29,58 @@
...
+#'
+#' # Subset to speed up example
+#' tse <- tse[, tse$subject %in% c("900", "934", "843", "875")]
+#'
+#' tse2 <- addBaselineDivergence(
+#'     tse,
+#'     group = "subject",
+#'     time_field = "time",
```
TuomasBorman: time.col

R/getBaselineDivergence.R
```diff
@@ -38,168 +91,218 @@
...
-                            n_dimred = NULL,
-                            ...){
+setMethod("addBaselineDivergence", signature = c(x = "SummarizedExperiment"),
+    function(
+        x,
+        time_field,
+        assay.type = "counts",
+        group = NULL,
+        name_divergence = "divergence",
+        name_timedifference = "time_diff",
```
TuomasBorman: `name.res`, `name.time`?

TuomasBorman: or `name` and` name.time`

TuomasBorman (changes requested):
This function is unnecessarily complex. We should aim to reuse existing functions whenever possible, as there is a lot of functionality here that is implemented only for this function, but it isn’t needed. Specifically, much of what this function does can already be handled by `mia::addDivergence`.

The only features missing from `addDivergence` that this function requires are:

a) Adding reference values to colData
b) Adding time difference results to colData

These could be integrated more simply without duplicating functionality.


`addBaselineDivergence`

-     (Optional) Set the first time point as the default reference.
1.     Calculate divergence values with `addDivergence`.
2.     Add time difference values to `colData`.

`addStepwiseDivergence`

1. Add reference sample information to `colData`.
2. Calculate divergence values with `addDivergence`.
3. Add time difference values to `colData`.

0xMuluh:
ndimred is longer used in `getStepwiseDivertgence`

R/deprecate.R
```diff
@@ -0,0 +1,52 @@
...
+    standardGeneric("getBaselineDivergence"))
+
+#' @rdname deprecate
+#' @export
+setMethod("getBaselineDivergence", signature = c(x = "ANY"), function(x, ...){
+    .Deprecated(msg = "'getBaselineDivergence' is deprecated. 
+                Use 'addBaselineDivergence' instead.")
+    addBaselineDivergence(x, ...)
+}
+)
```
TuomasBorman: I am wondering how established the functions are. Do we need to deprecate these and the parameter names? The feeling of mine is that the package has been in experimental stage as there has been only 2 functions. That would make our life easier if we can avoid deprecation.

@antagomir

R/getBaselineDivergence.R
```diff
@@ -38,168 +86,215 @@
...
+        assay.type = "counts",
+        group = NULL,
+        name = "divergence",
+        name.time = "time_diff",
+        method = "bray",
+        dimred = NULL,
+        ndimred = NULL,
+        dis.fun = vegan::vegdist,
+        baseline_sample = NULL,
+        ...){
```
TuomasBorman: Mixed parameter.name and parameter_name.

We use `getDivergence` from `mia` so `dis.fun` is not needed

TuomasBorman: Also:

Is it common to calculate divergences from the results of dimension reduction techniques like PCA? @antagomir 

If it is, then we should add this support to `mia::getDivergence()` (not here). The parameters could be passed with `...`. To reduce complexity of the function, this could be a hidden option.

R/getBaselineDivergence.R
```diff
@@ -38,168 +86,215 @@
...
-    # FIXME: reimplement the splitting so that we do not need intermediate variable like this
-    colData(x2)$tmp_group_for_groupwise_splitting <- NULL
-
+    
+    # FIXME: reimplement the splitting so that we do not need intermediate 
+    # variable like this
+    colData(x2)$group <- NULL
+    
     # Return
     return(x2)
```
TuomasBorman: Too complex.

R/getBaselineDivergence.R
```diff
@@ -38,168 +86,215 @@
...
         .check_altExp_present(altexp, x)
         x <- altExp(x, altexp)
     }
-    
-    if (is.null(colnames(x))) {
-        colnames(x) <- as.character(seq_len(ncol(x)))	
+    # assay.type
+    .check_assay_present(assay.type, x)
+    # time.col
+    temp <- .check_input(
```
TuomasBorman: Too long function. Should be splitted so that different steps are separated to unique internal functions.

R/getBaselineDivergence.R
```diff
@@ -227,18 +323,16 @@ getBaselineDivergence <- function(x,
     
     # Beta divergence from baseline info
     divergencevalues <- mia:::.calc_divergence(
-        cbind(mat, ref_mat), colnames(ref_mat), FUN = FUN, method = method)
+        cbind(mat, ref_mat), colnames(ref_mat), dis.fun = dis.fun, method = method)
     divergencevalues <- divergencevalues[seq_len(ncol(mat)), "value"]
-
+
```
TuomasBorman: Should be using getDivergence as we have the functionality already

R/getBaselineDivergence.R
```diff
@@ -38,168 +86,215 @@
...
-    # Global vars
-    is <- NULL
-
-    # If baseline is SE object then just ensure it has exactly one sample (well-defined baseline).
+.calculate_divergence_from_baseline <- function (x, baseline, time.col,
+                                                 name, 
+                                                 name.time,
+                                                 assay.type, dis.fun, method,
+                                                 dimred, ndimred) {
+
```
TuomasBorman: Indentation off probably

R/getBaselineDivergence.R
```diff
@@ -38,168 +86,215 @@
...
+    baseline_samples <- split(colData(x)[[baseline]], unfactor(colData(x)[[group]]))
+    correct <- lapply(baseline_samples, function(group){
+        # Get unique
+        group <- unique(group)
+        # It must be a single index or character specifying a column
+        res <- length(group) == 1 && (
+            (is.integer(group) && group >= 1 && group <= ncol(x)) ||
+                (is.character(group) && group %in% colnames(x)) )
+        return(res)
+    })
```
TuomasBorman: Lots of loops

R/getStepwiseDivergence.R
```diff
@@ -11,27 +11,25 @@
...
-#'   expect the input matrix as its first argument. With rows as samples 
-#'   and columns as features. (Default: \code{vegan::vegdist})
-#' @param method \code{Character scalar}. Used to calculate the distance. Method is
-#'   passed to the function that is specified by \code{FUN}. (Default: \code{"bray"})
-#' @param altexp \code{Character scalar} or \code{integer scalar}. Specifies the alternative experiment 
-#' containing the input data. (Default: \code{NULL})
-#' @param dimred \code{Character scalar} or \code{integer scalar}. indicates the reduced dimension
-#' result in `reducedDims` to use in the estimation. (Default: \code{NULL})
-#' @param n_dimred \code{Integer vector}. Specifies the dimensions to use if
+#' @param time_interval \code{Integer scalar}. Indicates the increment between
```
TuomasBorman: Should be time.interval

R/getStepwiseDivergence.R
```diff
@@ -42,205 +40,97 @@
...
+        x,
+        group=NULL,
+        time.col,
+        time_interval = 1,
+        name = "time_divergence",
+        name.time = "time_difference",
+        assay.type = "counts",
+        method="bray",
+        ndimred = NULL,
+        dimred = NULL,
```
TuomasBorman: List those parameters first that do not have any value. Here it should be: x, time.col, group = NULL

R/getStepwiseDivergence.R
```diff
@@ -42,205 +40,97 @@
...
+setMethod("addStepwiseDivergence", signature = c(x = "ANY"),
+    function(
+        x,
+        group=NULL,
+        time.col,
+        time_interval = 1,
+        name = "time_divergence",
+        name.time = "time_difference",
+        assay.type = "counts",
+        method="bray",
```
TuomasBorman: Add spaces around "=" ---> consistency

R/getStepwiseDivergence.R
```diff
@@ -42,205 +40,97 @@
...
+        )
+        ########################### INPUT CHECK END ############################
+        # Calculate values
+        x <- .add_previous_sample(x, group, time.col, time_interval )
+        res <- addDivergence(x, assay.type = assay.type, method = method, 
+                             reference = "previous_sample", 
+                             name = name, ndimred = ndimred, dimred = dimred, ...)
+        col_data <- colData(res)
+        colnames(col_data)[colnames(col_data) == "time_diff"] <- name.time
+        colData(res) <- col_data
```
TuomasBorman: Very nice!

R/getStepwiseDivergence.R
```diff
@@ -42,205 +40,97 @@
...
+    arrange(.data[[group]], .data[[time]]) %>%
+    group_by(.data[[group]]) %>%
+    # Lag time by 1 (previous time point)
+    mutate(previous_time = lag(.data[[time]], n = time_interval),  
+           # Lag sample name by 1
+           previous_sample = lag(sample, n = time_interval)) %>%  
+    ungroup() |> DataFrame()
+  rownames(df) <- df$sample
+  df[["time_diff"]] <- df[[time]] - df[["previous_time"]]
+  df <- df[ match(colnames(x), rownames(df)), ]
```
TuomasBorman: Works nicely!

R/getStepwiseDivergence.R
```diff
@@ -42,205 +40,97 @@
...
-
-        getStepwiseDivergence(x, ...)
+setMethod("addStepwiseDivergence", signature = c(x = "ANY"),
+    function(
+        x,
+        group=NULL,
+        time.col,
+        time_interval = 1,
+        name = "time_divergence",
+        name.time = "time_difference",
```
TuomasBorman: Not sure if `name` and `name.time` are the best names, although cannot think better... We can discuss about that

R/getBaselineDivergence.R
```diff
@@ -38,168 +86,215 @@
...
+        time.col,
+        assay.type = "counts",
+        group = NULL,
+        name = "divergence",
+        name.time = "time_diff",
+        method = "bray",
+        dimred = NULL,
+        ndimred = NULL,
+        dis.fun = vegan::vegdist,
+        baseline_sample = NULL,
```
TuomasBorman: reference might be better name for this as we use the same name in mia::getDivergence()

TuomasBorman (changes requested):
In the most recent code, I made some changes to minimize redundant code. The functionality of these functions are now ready, however, the documentation is not updated yet.

Moreover, we need to discuss about deprecation (do we need to deprecate or can we just remove and modify). Especially this is relevant in `get*` functions as they previously added values to `colData` but now they return only the values (as we have decided this naming scheme).

Moreover, `reducedDim()` is not now supported. If we want to support it, it needs to be incorporated to `mia::getDivergence()`. The idea of these miaTime divergence functions is only to add reference samples (and time difference) to `colData`, `mia::getDivergence()` does the calculation.

R/getStepwiseDivergence.R
TuomasBorman: I think these are ok because we have used `name` widely to specify a column name for resulted values

R/getBaselineDivergence.R
TuomasBorman: This will be taken into account in `mia::getDivergence()`

TuomasBorman: https://github.com/microbiome/mia/pull/653

TuomasBorman:
I think this PR is now ready. However, there is still one important thing to discuss. We have not deprecated the argument names. Moreover, previously `get*Divergence` functions returned `TreeSE` with calculated values in `colData`.

Now they `get*` functions return `DF` and `add*` return `TreeSE`. Because we will not remove the `get*` functions but just modify them to return `DF`, we cannot deprecate them.

Moreover, deprecating argument name causes extra work.

--> Can these arguments and functions be modified without deprecation? I think (but I do not know) that these functions are not that widely used yet, so we can modify them freely. Moreover, this package is not published yet and this has been more like a draft.

@antagomir

antagomir:
I agree that this can be changed now without deprecation, not yet in wide use.

R/getBaselineDivergence.R
antagomir: Sry for delays - I dont think it is very common and not necessarily so good practice to calculate divergences from the results of dimension reduction techniques like PCA. It could be sometimes useful to illustrate stuff in PCA space, so if this was already added then it can be there. Otherwise, possible to add that later if needs arise.

R/getStepwiseDivergence.R
antagomir: Ok to me.

Btw term "time" could be also more generic, like "sorter" (...) because the sorting variable could in principle be something else than time. But perhaps this is not so crucial now.

R/deprecate.R
antagomir: Yes, we can avoid that and just proceed.

antagomir:
Yes, these can be modified without deprecation. 

Is this now ready to merge & close?

antagomir:
Remember to check & update the vignettes and unit tests as well if necessary

R/getStepwiseDivergence.R
TuomasBorman: True. I think that for divergence values we have to have `name`. I was also wondering about `name = c("divergence", "time_diff")`, but then the input type differs between this and other functions. --> This is rather easy to change in the future, so we can fix this if we come up with better name.

One good point is that `name.time` is quite intuitive about the usage, `name.sorter` might not be

R/getBaselineDivergence.R
TuomasBorman: Yes, it is added in `getDivergence` as "hidden" option

TuomasBorman (approved):

PR #94, opened 2024-10-09
Short term change

0xMuluh:
Function to return short term change in a dataframe


0xMuluh:
@antagomir @TuomasBorman

TuomasBorman:
In holiday currently. I will check this early next week.

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+        }
+        # Check assay.type
+        mia:::.check_assay_present(assay.type, x)
+        if(!mia:::.is_a_bool(rarefy)){
+            stop("'rarefy' must be TRUE or FALSE.", call. = FALSE)
+        }
+        if(!mia:::.is_a_bool(compositional)){
+            stop("'compositional' must be TRUE or FALSE.", call. = FALSE)
+        }
+        # Ensure that the provided depth is valid
```
TuomasBorman: Instead, these input check functions should be imported in utils.R so that we can use them without mia:::

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+        if ( !is.null(depth) && depth > min(assay(x, assay.type)) ) {
```
TuomasBorman: min() --> min( na.rm = TRUE) so that it does not fail with NAs

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+            stop("Depth cannot be greater than the minimum number of counts in your data")
```
TuomasBorman: call. = FALSE

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+        } else if (rarefy == FALSE && compositional == FALSE) {
+            message("rarefy is set to FALSE, compositional==FALSE, using raw counts")
+            x <- x
+        } else if (rarefy == FALSE && compositional == TRUE) {
+            message("rarefy is set to FALSE, compositional==TRUE, using relative abundances")
+            x <- transformAssay(x, method = "relabundance", assay.type = assay.type)
+            assay.type <- "relabundance"
+        } else if (rarefy == TRUE && compositional == TRUE) {
+            stop("Both rarefy and compositional cannot be TRUE simultaneously")
+        }
```
TuomasBorman: This should be done outside the function

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+    
+    # Reshape data
+    assay_data <- reshape2::melt(assay_data)
+    assay_data <- as_tibble(assay_data) %>%
+        arrange(Var2) %>%
+        group_by(Var1) %>%
+        mutate(
+            time_lag = Var2 - lag(Var2), 
+            growth_diff = value - lag(value),
+            growth_rate = (value - lag(value)) / lag(value),
```
TuomasBorman: mia::meltSE should be used

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+            message("rarefy is set to FALSE, compositional==FALSE, using raw counts")
+            x <- x
+        } else if (rarefy == FALSE && compositional == TRUE) {
+            message("rarefy is set to FALSE, compositional==TRUE, using relative abundances")
+            x <- transformAssay(x, method = "relabundance", assay.type = assay.type)
+            assay.type <- "relabundance"
+        } else if (rarefy == TRUE && compositional == TRUE) {
+            stop("Both rarefy and compositional cannot be TRUE simultaneously")
+        }
+        ########################### Growth Metrics ############################
```
TuomasBorman: Too long function, should be splitted into smaller ones

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+    # Reshape data
+    assay_data <- reshape2::melt(assay_data)
+    assay_data <- as_tibble(assay_data) %>%
+        arrange(Var2) %>%
+        group_by(Var1) %>%
+        mutate(
+            time_lag = Var2 - lag(Var2), 
+            growth_diff = value - lag(value),
+            growth_rate = (value - lag(value)) / lag(value),
+            var_abund = (value - lag(value)) / time_lag
```
TuomasBorman: It seems that there is no way to control the grouping (these values are mixing the values from multiple instances

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+#' tse <- minimalgut
+#' 
+#' short_time_labels <- c("74.5h", "173h", "438h", "434h", "390h")
+#' 
+#' # Subset samples by Time_lable and StudyIdentifier
+#' tse <- tse[, !(colData(tse)$Time_label %in% short_time_labels)]
+#' tse <- tse[, (colData(tse)$StudyIdentifier == "Bioreactor A")]
+#' 
+#' # Plot short term change in abundance
+#' shortTermChange(tse, rarefy = TRUE, plot = TRUE) + ggtitle("Bioreactor A")
```
TuomasBorman: After this line, add NULL

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+            mutate(ismax = ifelse(growth_diff == max.growth, T, F))
+
+        grs.all$OTU <- gsub("_", " ", grs.all$OTU)
+        
+        grs.all$OTUabb <- toupper(abbreviate(grs.all$OTU,
+                                             minlength = 3,
+                                             method = "both.sides"
+        ))
+        grs.all$otu.time <- paste0(grs.all$OTUabb, " ", grs.all$time, "h")
+        if(plot){
```
TuomasBorman: I think plotting should be as separate function. Moeover, improve comments; what and why something is done

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+        } else if (rarefy == FALSE && compositional == TRUE) {
+            message("rarefy is set to FALSE, compositional==TRUE, using relative abundances")
+            x <- transformAssay(x, method = "relabundance", assay.type = assay.type)
+            assay.type <- "relabundance"
+        } else if (rarefy == TRUE && compositional == TRUE) {
+            stop("Both rarefy and compositional cannot be TRUE simultaneously")
+        }
+        ########################### Growth Metrics ############################
+        grwt <- .calculate_growth_metrics(x, assay.type = assay.type, ...)
+        #max growth
```
TuomasBorman: Add space after #

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+        }
+        return(grs.all)
+    
+    }
+)
+        ########################################################################
+# wrapper to calculate growth matrix
+.calculate_growth_metrics <- function(x, assay.type, ...) {
+    assay_data <- assay(x, assay.type)
+    time <- colData(x)[, "Time.hr"]
```
TuomasBorman: Do not hardcode time point column

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+        grs.all <- grs.all %>%
+            mutate(ismax = ifelse(growth_diff == max.growth, T, F))
+
+        grs.all$OTU <- gsub("_", " ", grs.all$OTU)
+        
+        grs.all$OTUabb <- toupper(abbreviate(grs.all$OTU,
+                                             minlength = 3,
+                                             method = "both.sides"
+        ))
+        grs.all$otu.time <- paste0(grs.all$OTUabb, " ", grs.all$time, "h")
```
TuomasBorman: Column names are not in unified format

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+        standardGeneric("shortTermChange"))
+
+#' @rdname shortTermChange
+#' @export
+#' @importFrom dplyr arrange as_tibble summarize
+#' @importFrom ggplot2 ggplot aes
+#' @importFrom ggrepel geom_text_repel
+#' @importFrom mia rarefyAssay transformAssay
+#' @importFrom SummarizedExperiment colData
+setMethod("shortTermChange", signature = c(x = "SummarizedExperiment"),
```
TuomasBorman: The naming should follow the scheme --> getShortTermChange

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+#' tse <- tse[, (colData(tse)$StudyIdentifier == "Bioreactor A")]
+#' 
+#' # Plot short term change in abundance
+#' shortTermChange(tse, rarefy = TRUE, plot = TRUE) + ggtitle("Bioreactor A")
+
+
+#' @rdname shortTermChange
+#' @export
+setGeneric("shortTermChange", signature = c("x"),
+    function(x,assay.type = "counts", rarefy = FALSE, compositional = FALSE,
```
TuomasBorman: Add space after `x,`

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+            x <- transformAssay(x, method = "relabundance", assay.type = assay.type)
+            assay.type <- "relabundance"
+        } else if (rarefy == TRUE && compositional == TRUE) {
+            stop("Both rarefy and compositional cannot be TRUE simultaneously")
+        }
+        ########################### Growth Metrics ############################
+        grwt <- .calculate_growth_metrics(x, assay.type = assay.type, ...)
+        #max growth
+        maxgrs <- grwt %>%
+            summarize(max.growth = max(growth_diff, na.rm = T))
```
TuomasBorman: Do not use T or F, use TRUE and FALSE

R/shortTermChange.R
```diff
@@ -0,0 +1,189 @@
...
+        grs.all <- merge(grwt, maxgrs, by = "OTU")
+        grs.all <- grs.all %>%
+            mutate(ismax = ifelse(growth_diff == max.growth, T, F))
+
+        grs.all$OTU <- gsub("_", " ", grs.all$OTU)
```
TuomasBorman: The taxa is not necessarily OTU which means operational taxonomy unit.

TuomasBorman (changes requested):
This PR is probably also not the most urgent ones

The calcultion part / core functionality is rather simple. The function should:

1. Get data with mia::meltSE (transformations sould be applied in prior to this function)
2. Calculate short term change based on `group` and `time.co`l
3. Return the data.frame

After this, user could plot the results if wanted

R/shortTermChange.R
antagomir: yes would be better so

antagomir:
Good to finalize this as proposed.

R/shortTermChange.R
antagomir: Also, you could write shorter like tse$Time_label

antagomir:
Example shows the following:

```
# Load time series data
#' data(minimalgut)
#' tse <- minimalgut
#' 
#' short_time_labels <- c("74.5h", "173h", "438h", "434h", "390h")
#' 
#' # Subset samples by Time_lable and StudyIdentifier
#' tse <- tse[, !(colData(tse)$Time_label %in% short_time_labels)]
#' tse <- tse[, (colData(tse)$StudyIdentifier == "Bioreactor A")]
#' 
#' # Plot short term change in abundance
#' shortTermChange(tse, rarefy = TRUE, plot = TRUE) + ggtitle("Bioreactor A")
```

Is this mean to be an analysis or visualization function?

The example could explain a bit more what this aims to show. I am not sure if I understand the motivation of this function from the manpage description.

It would be good to have a separate function, like getShortTermChange / addShortTermChange and then a separate function for visualization (unless it is easy to do with existing functions, like miaViz::plotSeries).

Is this linked to some miaTime issue? Issue #7 ?

R/shortTermChange.R
0xMuluh: done.

R/shortTermChange.R
0xMuluh: ACK.

R/shortTermChange.R
0xMuluh: ACK.

R/shortTermChange.R
0xMuluh: done.

R/shortTermChange.R
0xMuluh: ACK.

R/shortTermChange.R
0xMuluh: ACK.

R/shortTermChange.R
0xMuluh: ACK.

R/shortTermChange.R
0xMuluh: ACK.

R/shortTermChange.R
0xMuluh: ACK.

R/shortTermChange.R
0xMuluh: ACK.

R/shortTermChange.R
0xMuluh: done. pass time point columne as `time.col`

0xMuluh:
> #' 
> #' # Plot short term change in abundance
> #' shortTermChange(tse, rarefy = TRUE, plot = TRUE) + ggtitle("Bioreactor A")
> ```
> 
> Is this mean to be an analysis or visualization function?

original reference had plotting as part of the function. It is now removed and the function returns a dataframe with short term changes

> 
> Is this linked to some miaTime issue? Issue #7 ?

yes, syncomR consumed phyloseq object.

0xMuluh:
Moreover given the nature of the data frame results, I am not sure plotSeries is the right visualization method for it.  Suppose we somehow `addShortTermChange` using one of the tse slots(say maybe `reducedDims`),  it would still require it's own ggplot wrapper . 
I just checked but to be sure @TuomasBorman do we have miaViz function that takes dataframes?
For reference, this is how the plots from syncomR could look like linear or polarized: 

Rplotstc
Rplot01stc

and of course we can also discuss how we want to represent our own plots.

R/getShortTermChange.R
```diff
@@ -0,0 +1,146 @@
...
+    } else if (rarefy == FALSE && compositional == FALSE) {
+        message("rarefy is set to FALSE, compositional==FALSE, using raw counts")
+        x <- x
+    } else if (rarefy == FALSE && compositional == TRUE) {
+        message("rarefy is set to FALSE, compositional==TRUE, using relative abundances")
+        x <- transformAssay(x, method = "relabundance", assay.type = assay.type)
+        assay.type <- "relabundance"
+    } else if (rarefy == TRUE && compositional == TRUE) {
+        stop("Both rarefy and compositional cannot be TRUE simultaneously", call. = FALSE)
+    }
```
TuomasBorman: I think these should be done outside the function

R/getShortTermChange.R
```diff
@@ -0,0 +1,146 @@
...
+            growth_rate = (!!sym(assay.type) - lag(!!sym(assay.type))) / lag(!!sym(assay.type)),
+            var_abund = (!!sym(assay.type) - lag(!!sym(assay.type))) / time_lag
+        )
+    return(assay_data)
+}
+
+.clean_growth_metrics <- function(grwt, time.col = NULL, ...) {
+    # Calculate max growth
+    maxgrs <- grwt %>%
+        summarize(max.growth = max(growth_diff, na.rm = TRUE))
```
TuomasBorman: The df has now column names with three different styles: "column_name" , "columname" and "column,name" --> use the first one

R/getShortTermChange.R
```diff
@@ -0,0 +1,146 @@
...
+        # Check assay.type
+        .check_assay_present(assay.type, x)
+        if(!.is_a_bool(rarefy)){
+            stop("'rarefy' must be TRUE or FALSE.", call. = FALSE)
+        }
+        if(!.is_a_bool(compositional)){
+            stop("'compositional' must be TRUE or FALSE.", call. = FALSE)
+        }
+        # Ensure that the provided depth is valid
+        if ( !is.null(depth) && depth > min(assay(x, assay.type), na.rm = TRUE) ) {
```
TuomasBorman: You are using styles `if (something)` and `if(something)`. Good to be consistent

R/getShortTermChange.R
```diff
@@ -0,0 +1,146 @@
...
+                                      rarefy, compositional, depth, ...) {
+    ############################ Data Preparation ##############################
+    # Initialize the filtered object based on rarefy and compositional arguments
+    if (rarefy == TRUE && compositional == FALSE) {
+        message("rarefy is set to TRUE, calculating short term change using counts")
+        x <- rarefyAssay(x, assay.type = assay.type, depth = depth)
+        assay.type <- "subsampled"
+    } else if (rarefy == FALSE && compositional == FALSE) {
+        message("rarefy is set to FALSE, compositional==FALSE, using raw counts")
+        x <- x
```
TuomasBorman: This does not do anything

R/getShortTermChange.R
```diff
@@ -0,0 +1,146 @@
...
+    # Calculate max growth
+    maxgrs <- grwt %>%
+        summarize(max.growth = max(growth_diff, na.rm = TRUE))
+    # Merge growth data with max growth
+    grs.all <- merge(grwt, maxgrs, by = "FeatureID")
+    # Add 'ismax' column indicating if the growth is the maximum
+    grs.all <- grs.all %>%
+        mutate(ismax = ifelse(growth_diff == max.growth, TRUE, FALSE))
+    # Clean and abbreviate FeatureID names
+    grs.all$FeatureID <- gsub("_", " ", grs.all$FeatureID)
```
TuomasBorman: What if the rownames has these underscores? Use cannot then match them directly with the original data

TuomasBorman (changes requested):
Looks nice

TuomasBorman:
> Moreover given the nature of the data frame results, I am not sure plotSeries is the right visualization method for it. Suppose we somehow `addShortTermChange` using one of the tse slots(say maybe `reducedDims`), it would still require it's own ggplot wrapper . I just checked but to be sure @TuomasBorman do we have miaViz function that takes dataframes? For reference, this is how the plots from syncomR could look like linear or polarized:
> 
> Rplotstc Rplot01stc
> 
> and of course we can also discuss how we want to represent our own plots.

There is no that kind of restriction that miaViz cannot have functions that take df as input. In fact, loadings plotting function can take df as input,

This outputs currently data.frame and I think it makes sense. reducedDims is not correct because this does not reduce dimensionality. This could be stored to metadata and then we could have a plotting function for TreeSE and df.

R/getShortTermChange.R
antagomir: Rarification is part of calculating beta diversity (ie. dissimilarity), whereever that happens.

R/getShortTermChange.R
antagomir: or even col_name ?

R/getShortTermChange.R
antagomir: Yep, in most cases we would not like to modify feature IDs in the functions.

antagomir:
Or can it be part of colData? Could be simpler. This is a measure that is available on a per-sample basis. Except that some samples may have NAs if they are missing the comparison.

R/getShortTermChange.R
TuomasBorman: But this could be done outside the function to avoid duplicated calculations. We have also other functions where relative abundances are meant to be used, but user is advised to do it before running the function.

antagomir: Yes I agree. I meant that if beta diversity calculation is in some cases part of the procedure within the function then the rarefaction arguments should be passed rather than reimplemented in the function. But now I realize that we do not have rarified version of the beta diversity calculation - that was problematic to implement because rarefaction can be followed by transformation before dissimilarity calculation and in general we do not like mix up these too much.. although I still may think that perhaps it would be rather handy to support rarefaction in `getDissimilarity` but that's another question..

TuomasBorman:
> Or can it be part of colData? Could be simpler. This is a measure that is available on a per-sample basis. Except that some samples may have NAs if they are missing the comparison.

That could be possible but then columns would denote features. It might be confusing and there might be lots of features to add in the columns

R/getShortTermChange.R
TuomasBorman: `getDissimilarity` actually nowadays support rarefaction via using `vegan::avgdist`

https://microbiome.github.io/mia/reference/getDissimilarity.html

antagomir:
> > Or can it be part of colData? Could be simpler. This is a measure that is available on a per-sample basis. Except that some samples may have NAs if they are missing the comparison.
> 
> That could be possible but then columns would denote features. It might be confusing and there might be lots of features to add in the columns

Yep, not so optimal.

antagomir:
"This function essentially calculates the change in abundance over time for a microbe by abundancet2/ abundancet1."

We already have [getStepwiseDivergence](https://microbiome.github.io/miaTime/reference/addStepwiseDivergence.html)?

`getStepwiseDivergence` calculates x(t + d) - x(t) for abundance x at time points t+d vs. t any user defined interval d along the time series.

`getShortTermChange` calculates this per taxa. One can generalize and use it for a more general interval x(t+d)/x(t), analogous to `getStepwiseDivergence`. 

`getShortTermChange` suggests to do x(t+d)/x(t) instead of x(t+d)-x(t). But both are available and the choice may also depend on whether the data is logarithmic. log(a/b) = log(a)-log(b) although zeroes may cause trouble. It could be enough to implement the difference "x-y" and the rest ("x/y") can be always dealt (by the user?) with log/exp operations the data if necessary.

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+
+################################ HELP FUNCTIONS ################################
+
+# wrapper to calculate growth matrix
+.calculate_growth_metrics <- function(x, assay.type, time.col = NULL, ...) {
+
+    # Reshape data and calculate growth metrics
+    assay_data <- meltSE(x, assay.type = assay.type, 
+                         add.col = time.col, row.name = "Feature_ID")
+    assay_data <- assay_data %>%
```
TuomasBorman: Use |> operators. They are used in elsewhere

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+        return(grs.all)
+    }
+)
+
+#' @rdname addShortTermChange
+#' @export
+setMethod("addShortTermChange", signature = c(x = "SummarizedExperiment"),
+          function(x, assay.type = "counts", name = "short_term_change", ...){
+              # Calculate short term change
+              res <- getShortTermChange(x, ...)
```
TuomasBorman: Check BiocHCeck::BiocHeck() at least indentation is off

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
+#' object with these results in its \code{metadata}.
+#' 
+#' @details This approach is used by Wisnoski NI and colleagues
+#'          \url{https://github.com/nwisnoski/ul-seedbank}. Their approach is based on
+#'          the following calculation log(present abundance/past abundance).
+#'          Also a compositional version using relative abundance similar to
+#'          Brian WJi, Sheth R et al
+#'          \url{https://www.nature.com/articles/s41564-020-0685-1} can be used.
+#'          This approach is useful for identifying short term growth behaviors of taxa.
```
TuomasBorman: Remove this indentation

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+#' 
+#' @examples
+#' 
+#' # Load time series data
+#' data(minimalgut)
+#' tse <- minimalgut
+#' 
+#' short_time_labels <- c("74.5h", "173h", "438h", "434h", "390h")
+#' 
+#' # Subset samples by Time_label and StudyIdentifier
```
TuomasBorman: Instead of this comment, say what it is the goal: get specified time points from certain bioreactor

R/getShortTermChange.R
TuomasBorman: And also from above

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+#' @param ... additional arguments.
+#' 
+#' 
+#' @return \code{getShortTermChange} returns \code{DataFrame} object containing 
+#' the short term change in abundance over time for a microbe. 
+#' \code{addShortTermChange}, on the other hand, returns a
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
+#' object with these results in its \code{metadata}.
+#' 
+#' @details This approach is used by Wisnoski NI and colleagues
```
TuomasBorman: You could use math notation, it would be easier to read. Also the calculated values include other measurements than this

TuomasBorman: You could describe them also

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+#' tse <- minimalgut
+#' 
+#' short_time_labels <- c("74.5h", "173h", "438h", "434h", "390h")
+#' 
+#' # Subset samples by Time_label and StudyIdentifier
+#' tse <- tse[, !(tse$Time_label %in% short_time_labels)]
+#' tse <- tse[, (tse$StudyIdentifier == "Bioreactor A")]
+#' 
+#' # Get short term change
+#' # Case of rarefying counts
```
TuomasBorman: This does not rarefy counts. Are the comments in wrong place?

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+#' getShortTermChange(tse, assay.type = "subsampled", time.col = "Time.hr")
+NULL
+
+#' @rdname addShortTermChange
+#' @export
+#' @importFrom dplyr arrange as_tibble summarize "%>%"
+#' @importFrom ggplot2 ggplot aes
+#' @importFrom ggrepel geom_text_repel
+#' @importFrom mia rarefyAssay transformAssay
+#' @importFrom SummarizedExperiment colData
```
TuomasBorman: Add these above the function that utilizes these. Easier to remove when we do not need them anymore

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+#' tse <- rarefyAssay(tse, assay.type = "counts")
+#' getShortTermChange(tse, assay.type = "subsampled", time.col = "Time.hr")
+NULL
+
+#' @rdname addShortTermChange
+#' @export
+#' @importFrom dplyr arrange as_tibble summarize "%>%"
+#' @importFrom ggplot2 ggplot aes
+#' @importFrom ggrepel geom_text_repel
+#' @importFrom mia rarefyAssay transformAssay
```
TuomasBorman: These are not used

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+              x <- .add_values_to_metadata(x, name, res, ...)
+              return(x)
+          }
+)
+
+################################ HELP FUNCTIONS ################################
+
+# wrapper to calculate growth matrix
+.calculate_growth_metrics <- function(x, assay.type, time.col = NULL, ...) {
+
```
TuomasBorman: Remove this empty line

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+              res <- getShortTermChange(x, ...)
+              # Add to metadata
+              x <- .add_values_to_metadata(x, name, res, ...)
+              return(x)
+          }
+)
+
+################################ HELP FUNCTIONS ################################
+
+# wrapper to calculate growth matrix
```
TuomasBorman: Start comments with capital

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+.calculate_growth_metrics <- function(x, assay.type, time.col = NULL, ...) {
```
TuomasBorman: time.col is always needed? Then add it to exported function. Also it is not checked that it can be found from colData(). (meltSE checks that but the error message says about add.col)

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+        grs.all <- .clean_growth_metrics(grwt, ...)
+        return(grs.all)
+    }
+)
+
+#' @rdname addShortTermChange
+#' @export
+setMethod("addShortTermChange", signature = c(x = "SummarizedExperiment"),
+          function(x, assay.type = "counts", name = "short_term_change", ...){
+              # Calculate short term change
```
TuomasBorman: To support altexp, you could add .get_and_check_altExp() similarly to divergence functions

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+    # Add 'ismax' column indicating if the growth is the maximum
+    grs.all <- grs.all %>%
+        mutate(ismax = ifelse(growth_diff == max_growth, TRUE, FALSE))
+    # Clean and abbreviate FeatureID names
+    grs.all$Feature_IDabb <- toupper(abbreviate(grs.all$Feature_ID, 
+                                               minlength = 3, 
+                                               method = "both.sides"))
+    # Create 'Feature.time' column combining abbreviation and time information
+    grs.all$Feature_time <- paste0(grs.all$Feature_IDabb, " ", 
+                                   grs.all[[time.col]], "h")
```
TuomasBorman: Is this line needed? The info is already in the table. Moreover, the unit can be other than hours

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+
+.clean_growth_metrics <- function(grwt, time.col = NULL, ...) {
+    # Calculate max growth
+    maxgrs <- grwt %>%
+        summarize(max_growth = max(growth_diff, na.rm = TRUE))
+    # Merge growth data with max growth
+    grs.all <- merge(grwt, maxgrs, by = "Feature_ID")
+    # Add 'ismax' column indicating if the growth is the maximum
+    grs.all <- grs.all %>%
+        mutate(ismax = ifelse(growth_diff == max_growth, TRUE, FALSE))
```
TuomasBorman: This column is in format "ismax" while others are in "is_max". Also it would be neater solution to just use `growth_diff == max_growth` as  it already returns boolean vector

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+        )
+    return(assay_data)
+}
+
+.clean_growth_metrics <- function(grwt, time.col = NULL, ...) {
+    # Calculate max growth
+    maxgrs <- grwt %>%
+        summarize(max_growth = max(growth_diff, na.rm = TRUE))
+    # Merge growth data with max growth
+    grs.all <- merge(grwt, maxgrs, by = "Feature_ID")
```
TuomasBorman: for variables use grs_all. grs.all is reserved for parameters

tests/testthat/test-getShortTermChange.R
```diff
@@ -0,0 +1,25 @@
...
+    tse_filtered <- tse_filtered[, (tse_filtered$StudyIdentifier == "Bioreactor A")]
+    
+    expect_true(all(!(tse_filtered$Time_label %in% short_time_labels)))
+    
+    result <- getShortTermChange(tse_filtered, time.col = "Time.hr")
+    # Expected output is a dataframe
+    expect_true(is.data.frame(result))  
+    expect_true("growth_diff" %in% colnames(result))
+    # Test some expected properties (e.g., that growth_diff isn't all NAs)
+    expect_false(all(is.na(result$growth_diff)))
```
TuomasBorman: The values should be also checked

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+
+#' @rdname addShortTermChange
+#' @export
+#' @importFrom dplyr arrange as_tibble summarize "%>%"
+#' @importFrom ggplot2 ggplot aes
+#' @importFrom ggrepel geom_text_repel
+#' @importFrom mia rarefyAssay transformAssay
+#' @importFrom SummarizedExperiment colData
+setMethod("getShortTermChange", signature = c(x = "SummarizedExperiment"),
+    function(x, assay.type = "counts", ...){
```
TuomasBorman: I think the function should also take into account the grouping, i.e., from which patient/bioreactor the data is coming

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+    assay_data <- assay_data %>%
+        arrange( !!sym(time.col) ) %>%
+        group_by(Feature_ID) %>%
+        mutate(
+            time_lag = !!sym(time.col) - lag( !!sym(time.col) ), 
+            growth_diff =!!sym(assay.type) - lag(!!sym(assay.type)),
+            growth_rate = (!!sym(assay.type) - lag(!!sym(assay.type))) / lag(!!sym(assay.type)),
+            var_abund = (!!sym(assay.type) - lag(!!sym(assay.type))) / time_lag
+        )
+    return(assay_data)
```
TuomasBorman: I think this should take into account duplicated values. I tink you could give warning first and then try something like this

```
df <- meltSE(tse, add.col = TRUE)
df |>
    group_by(StudyIdentifier, FeatureID, Time.hr) |>
    mutate(mean_abundance_for_timepoint = mean(counts, na.rm = TRUE)) |>
    distinct(StudyIdentifier, FeatureID, Time.hr, .keep_all = TRUE) |>
    ungroup() |>
    arrange(StudyIdentifier, FeatureID, Time.hr) |>
    group_by(StudyIdentifier, FeatureID) |>
    mutate(growth_diff = mean_abundance_for_timepoint - lag(mean_abundance_for_timepoint)) |>
    ungroup()
```

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+        summarize(max_growth = max(growth_diff, na.rm = TRUE))
+    # Merge growth data with max growth
+    grs.all <- merge(grwt, maxgrs, by = "Feature_ID")
+    # Add 'ismax' column indicating if the growth is the maximum
+    grs.all <- grs.all %>%
+        mutate(ismax = ifelse(growth_diff == max_growth, TRUE, FALSE))
+    # Clean and abbreviate FeatureID names
+    grs.all$Feature_IDabb <- toupper(abbreviate(grs.all$Feature_ID, 
+                                               minlength = 3, 
+                                               method = "both.sides"))
```
TuomasBorman: These feature names come directly from rownames()? Then I think they should be kept unchanged

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+        if (nrow(x) == 0L){
+            stop("No data available in `x` ('x' has nrow(x) == 0L.)",
+                 call. = FALSE)
+        }
+        # Check assay.type
+        .check_assay_present(assay.type, x)
+        ########################### Growth Metrics ############################
+        grwt <- .calculate_growth_metrics(x, assay.type = assay.type, ...)
+        # Clean and format growth metrics
+        grs.all <- .clean_growth_metrics(grwt, ...)
```
TuomasBorman: Not sure if these are needed, as these are quite basic metric that can be easily calculated

TuomasBorman (changes requested):
Looks good, some suggestions

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
+#' @title Short term Changes in Abundance
```
antagomir: Capitalize also "term"

R/getShortTermChange.R
```diff
@@ -0,0 +1,128 @@
...
+#'
+#' @description Calculates short term changes in abundance of taxa
+#'              using temporal Abundance data.
```
antagomir: "Abundance" -> "abundance"

antagomir:
I somehow missed what was the motivation for this function?

TuomasBorman:
> I somehow missed what was the motivation for this function?

Related to this issue: https://github.com/microbiome/miaTime/issues/7

TuomasBorman (approved):

antagomir (approved):

antagomir:
Seems good.

Issue #83, opened 2024-05-31
Conditionally & persistently rare taxa


antagomir:
In microbiome time series analyses we have the definitions of  abundant taxa, conditionally rare taxa, persistently rare taxa and other rare taxa (see e.g. the two refs below). 

Conditionally rare taxa are defined for instance as taxa with a maximum relative abundance at least N times higher than their minimum value. 

Persistently rare taxa are taxa whose maximum relative abundance never exceeds X times greater than the minimum. 

https://doi.org/10.1093%2Ffemsec%2Ffix126
https://doi.org/10.1128/mbio.01371-14

It could be helpful to have a function or example showing how to fetch these taxa sets from microbiome time series.

ginkgozh:
This article categorizes microorganisms based on threshold values：always AT (AAT) with a relative abundance of ≥1% in the dataset; conditionally AT (CAT) with ≥1% relative abundance in some samples, but never <0.01%; always RT (ART) with <0.01% relative abundance in all samples; conditionally RT (CRT) with <1% relative abundance in all initial samples, but <0.01% in some sample; moderate taxa (MT) with between 0.01% and 1% relative abundances in all data; and conditionally abundant and rare taxa (CRAT) whose relative abundances ranged from rare values of <0.01% to abundant values of ≥1%. As also previously described, CRAT, CAT, and AAT were collectively referred to as abundant taxa (AT), while ART and CRT comprised rare taxa (RT).（https://www.sciencedirect.com/science/article/pii/S0013935123028347#sec2）.

 However,` I am currently facing the same issue and cannot find the R code to classify them, nor do I know how to calculate their diversity indices after classification. If you have already solved this problem, please contact me (email: ginkgozh@163.com). Below is the only information I have obtained, the R code to classify them.  

`otu <- read.table("C:/Users/heng/Desktop/micro/BF.txt", header = TRUE, row.names = 1, sep = "\t")

#（i）稀有类群（rare taxa，RT），在所有样本中丰度均 ≤0.1% 的 OTU
otu_ART <- otu[apply(otu, 1, function(x) max(x) <= 0.0001), ]
#（ii）丰富类群（abundant taxa，AT），在所有样本中丰度均 ≥1% 的 OTU
otu_AAT <- otu[apply(otu, 1, function(x) min(x) >= 0.01), ]
#（iii）中间类群（moderate taxa，MT），在所有样本中丰度均 >0.1% 且 <1% 的 OTU
[...]

Issue #30, opened 2022-02-07
split-apply-combine


antagomir:
Many real applications require calculating time series analyses per subject. In fact this is more generic problem that can be solved on the same go for any grouping (not just subject, and not just time series).

Two cases:
1. For the full TreeSE object
2. For colData only (adding back to TreeSE; can be faster, if only colData is manipulated)

This is needed in many other functions.

Design and implement a general solution that takes as input 
- the SE object
- the grouping variable
- name of the function to apply
.. and then returns a new SE object where that function has been applied per group and combined back into a single SE object

Perhaps two functions: one for each case described above.

I am tagging @TuomasBorman here, he might find this interesting as well and could join the discussions. @YagmurSimsekk could give it the first try.

antagomir:
Functions `mia::splitOn` and `mia::unsplitOn` might help here now. This may require some more thinking.

antagomir:
This function could go to mia as well as it is so fundamental, not sure which package would be best. This doesnt necessarily add dependencies.

@TuomasBorman what's your view: mia or miaTime?

TuomasBorman:
Splitting TreeSE into a list of TreeSEs is not memory efficient. Instead, I think we should take this grouping into account internally in the methods that we develop (`group` parameter). Then we can use `dplyr` or loop through the subjects, but it is still faster because we are not copying the whole structure of TreeSE

There might be room for split-apply-combine (mia would be the right place if we decide to implement this as this is more general tool), but in practice this would only remove couple lines of code (and potentially make the functions harder to understand).

```
tse_list <- splitOn(tse, by = 2, group = "subject")
tse_list <- lapply(tse_list, function(x) doSomething(x, param1 = TRUE, param2 = FALSE))
tse <- mergeSEs(tse_list)
```
vs

`tse <- splitCombine(tse, by = 2, group = "subject", FUN = doSomething, param1 = TRUE, param2 = FALSE)`

antagomir:
True. Let us now implement such functionality now but it would be very useful to have an example in miaTime vignette. 

There is section "Operations by unit" already in the [miaTime vignette](https://microbiome.github.io/miaTime/articles/manipulation.html). We could add the above procedure there for a future reference.  @Daenarys8 could you suggest/add an example?

Or we could add it in OMA but miaTime is not in Bioconductor yet, and might be generally easier to first accumulate material in the package vignette, then later see if there is space for an extra chapter in OMA and do that properly at one go.

[...]

PR #98, opened 2024-11-14
conditionally and persisitently rare taxa

0xMuluh:
ping #83


R/addMetrics.R
```diff
@@ -0,0 +1,101 @@
...
+
+        # add results to metadata
+        x <- do.call(cbind, results_list)
+        
+        return(x)
+    }
+)
+
+# Internal function to calculate Conditional Rare Taxa (CRT)
+.calculateCRT <- function(data, threshold = 2) {
```
TuomasBorman: This should be .calculate_crt based on the naming scheme

R/addMetrics.R
```diff
@@ -0,0 +1,101 @@
...
+    # Extract the counts/abundance data
+    counts <- assay(data)
```
TuomasBorman: Shoudld have assay.type argument

R/addMetrics.R
```diff
@@ -0,0 +1,101 @@
...
+        # Check if 'split.by' exists in colData, otherwise do not split
+        if (!is.null(split.by) && !split.by %in% colnames(colData(x))) {
+            stop("The 'split.by' column does not exist in colData(x).", 
+                 call. = FALSE)
+        }
+
+        ############################## Calculate metrics #######################
+        # Split the data by subject (or other grouping variable) if required
+        if (!is.null(split.by)) {
+            split_data <- split(x, f = colData(x)[[split.by]])
```
TuomasBorman: Should this be splitOn

R/addMetrics.R
```diff
@@ -0,0 +1,101 @@
...
+                }
+            }
+
+            # Add the calculated metrics to the metadata (colData or rowData)
+            for (metric in names(metric_results)) {
+                colData(sub_data)[[metric]] <- metric_results[[metric]]
+            }
+            
+            return(sub_data)
+        })
```
TuomasBorman: It would be more efficient to:

0. Get matrix
1. Get the list which has vectors as elements. Each vector represent members of each group
2. Loop through this list
3. Select the group members from the matrix
4. Calculate values
5. Combine all the results and sort values based on sample names
6. Add to colData (in add* method)

TuomasBorman (changes requested):

R/addMetrics.R
0xMuluh: I am not convinced the right location should be colData as the interest is taxa related.

TuomasBorman: True, I followed the code and did not think about it. As it these features are global (for whole dataset), colData is not correct place indeed

I think the output could be similar to getPrevalent and getRare, i.e., vector

antagomir: Hmm.. yes. If we have a single subject/group with time series, then this is a vector over features (or rowData) - one measure of persistence or rarity for each feature and rowData would be a suitable location.

If we have multiple subjects/groups with time series, then this would be features x groups matrix.

At simplest it could be implemented for just a single-group case, and rest left for the user. 

If we support multi-group (multi-subject) analyses, then that is a more general design question and linked to #30

TuomasBorman: I think one vector could be calculated for whole dataset for now. As you mentioned, the multi-group thing is more general question, and is affecting also the mia functions

antagomir: Yes - note that this can only be calculated in the case where the whole data set has some continuous variable that extends all samples (like is the case when there is time dimension for a single subject). If we have multiple groups then this is not defined for the whole data set.

TuomasBorman: I might have missed something, but I cannot see direct linkage of time series to this concept of conditionally/persistent rare/abundant taxa. It seems that this is more general concept?

Based on the following article, I summarized the concept

Sizhong Yang, Matthias Winkel, Dirk Wagner, Susanne Liebner, Community structure of rare methanogenic archaea: insight from a single functional group, FEMS Microbiology Ecology, Volume 93, Issue 11, November 2017, fix126, https://doi.org/10.1093/femsec/fix126

Rare taxa:

- Average relative abundance below 1% accross all samples

Abundant taxa:

- Average relative abundance above 1% accross all samples

Conditionally rare taxa:

> microbial taxa that are typically in low abundance in one locality but occasionally become prevalent in at least one other sample

- CRT was assigned if an OTU exhibits a maximum relative abundance at least 100 times higher than its minimum value (max:min > 100)

Permanently rare taxa:

- If the (max/min) ratio is less than 5, that is the abundance rarely fluctuates among samples, it will be grouped into the permanently rare taxa (PER)

Rare and abundant taxa can be already calculated by mia functions:

`getRare(tse, assay.type = "relabundance", detection = 1/100)`
`getPrevalent(tse, assay.type = "relabundance", detection = 1/100)`


I see that these conditionally rare and permanently rare taxa are very close to `getRare()` and `getPrevalent()`. As they do not necessarily deal with time series, should these methods be in mia instead: `getConditionallyRare()` & `getPermanentlyRare()`

antagomir: Yep, I mixed it up, no time series needed. 

CRT seems like a bit strange concept in the end. I am not sure if there should be much difference whether the minimum count is 1, 2, or 3 reads but this can make a huge difference in the max:min calculation. But yes, it might be relevant to recognize taxa that have vast variation. Variance would not capture all these cases, if the blooms cover only few samples. Shouldn't this be called conditionally abundant, rather than conditionally rare..?

PER: is the rarity also evaluated, or just max/min ratio? some taxa might be abundant and prevalent, while having a low max/min ratio. At least in principle. Then it would be misleading to call them PER. The interesting PER are taxa that might be prevalent but always stay under a certain threshold and never bloom.

We could add the methods as suggested and it could go to mia. If this could be made into a more generic index function (like addAlpha for samples), that could help to simplify maintenance and adding possible other measures later. Would that be feasible..?

On the other if this seem time-consuming to finalize, it could be left for later.

TuomasBorman:
I read this article that Leo shared in the issue: https://journals.asm.org/doi/10.1128/mbio.01371-14

In this article, conditionally rare taxa is determined based on coefficient of bimodality. They determined CRT for each system from where multiple time points were collected.

How should we proceed with these? CRT is determined with two different ways.

Proposal:

1.

In mia, introduce new funtions:

`getConditionallyRare()` and `getPermanentlyRare()`.

These functions would return a vector of taxa (only those ones that are classified as CRT or PRT). Similar to `getRare()` and `getPrevalent()`.

These would follow the implementation introduced in this article

Sizhong Yang, Matthias Winkel, Dirk Wagner, Susanne Liebner, Community structure of rare methanogenic archaea: insight from a single functional group, FEMS Microbiology Ecology, Volume 93, Issue 11, November 2017, fix126, https://doi.org/10.1093/femsec/fix126

Conditionally rare:

(max:min > 100), where max and min are maximum and minimum abundances

Permantently rare:

(max:min < 5)

2.

In mia, implement `getPrevelanceClass()` function that returns a class for each taxa (prevalent, CRT, PRT or rare).

3.

In miaTime, implement `getBimodalityCoefficient()`. For each system/group/patient and taxa, return bimodality coefficient.

The method would return either a vector length nrow or data.frame with dimensions nrow*N_system

This method would follow this implementation:

Shade A, Jones SE, Caporaso JGHandelsman J, Knight RFierer N, Gilbert JA2014.Conditionally Rare Taxa Disproportionately Contribute to Temporal Changes in Microbial Diversity. mBio5:10.1128/mbio.01371-14.https://doi.org/10.1128/mbio.01371-14

@antagomir

antagomir:
There are different multi/bimodality indices available, here one implementation:
https://cran.r-project.org/web/packages/BimodalIndex/BimodalIndex.pdf

antagomir:
Interesting, I had not paid attention to the multiple definitions. 

Definitions in (1-2) are somehow arbitrary, and relatively easy calculate manually in whatever desirable way. I am hesitating if we need full methods for these in the end, before there is a clear need / more use cases. On the other hand this PR is already half-way..

Bimodalities are interesting, too. I'm good with that if it is quick to add. It could make sense to check what uni/bi/multimodality indices are available in R. There are a couple of different methods.

TuomasBorman:
1-2) I think this PR is not suitable for implementing these since I think mia is the correct place. However, these should be quite easy to implement.

```
getConditionallyRare <- function(x, threshold = 100, ...){
    res <- .calculate_max_min()
    res <- names(res)[ res > threshold ]
    return(res)
}

getPermanentlyRare <- function(x, threshold = 5, ...){
    res <- .calculate_max_min()
    res <- names(res)[ res < threshold ]
    return(res)
}
```

3) This should be also fast to add especially if we decide to utilize these existing packages. It would make sense to use them but probably have to check what options there are (if there are some other package with broader scope, and we could use it elsewhere also)

antagomir:
Seems good. Only question I have is whether the added benefit from having these will be higher than the maintenance burden as these are relatively specialized functions. I could have thought about it earlier. But if these are fast to add, then it would be good to just PR them to mia and close the case.

antagomir:
Re: (3) I would do that when someone has need in a real project.

TuomasBorman:
1-2) That is true. This is especially problematic if there are more than these two ways to define conditionally rare and persistently rare taxa. I don't know if it is common to determine these CRT and PRT, and are these the most common definitions.

However to point out, in the issue, there was one external person interested in this activity. If these are the most common definitions for CRT & PRT, it would be nice to add features that community needs.

3. I agree that this bimodality thing can be implemented in the future if needed

TuomasBorman:
Closing. Then described things are dealt in other PRs:

- https://github.com/microbiome/miaTime/pull/103
- https://github.com/microbiome/mia/pull/664

Issue #110, opened 2025-04-20
New demo data


antagomir:
This recent paper (of ours) has seasonal time series:
https://www.nature.com/articles/s41467-025-56014-6

Data is available in Zenodo, including tse.Rds object (see Code availabilty statement).

This is a useful recent study including short time series.

Add and document demo data set in the miaTime package (similar to other existing demo data sets).

PR #111, opened 2025-04-20
add Crest study dataset

0xMuluh:
ping: #110


R/miaTime.R
```diff
@@ -138,3 +138,50 @@ NULL
...
+#' matched for sex, BMI, and age.
+#' Indian subjects were grouped into four diet groups (~20 subjects per group), 
+#' and fecal samples were collected across three seasonal time points.
+#'
+#' The microbiota profiling was performed using HITChip microarray analysis 
+#' (in duplicate), qPCR (in triplicate with eight-point standard curves), and 
+#' LC-HRMS and HPLC metabolite profiling with internal standards.
+#'
+#' Column metadata includes diet group assignment, sampling season, sex, BMI, 
+#' age, and questionnaire-based lifestyle metadata.
```
TuomasBorman, line 202: Even though this data comes from publication, we could try to polish it little bit:

1. Harmonize colData columns
Currently, they are with different styles (dry.fish, Hentak..fermentd.fish.paste..boiled..)

--> All lower case and "_" between words

2. Instead of y/n, use TRUE/FALSE (more explicit)

R/miaTime.R
```diff
@@ -138,3 +138,50 @@ NULL
 #' @keywords data
 #'
 NULL
+
+#' @title CrestStudyData
```
TuomasBorman: From where this name comes from?

R/miaTime.R
```diff
@@ -138,3 +138,50 @@ NULL
...
+#'
+#' The microbiota profiling was performed using HITChip microarray analysis 
+#' (in duplicate), qPCR (in triplicate with eight-point standard curves), and 
+#' LC-HRMS and HPLC metabolite profiling with internal standards.
+#'
+#' Column metadata includes diet group assignment, sampling season, sex, BMI, 
+#' age, and questionnaire-based lifestyle metadata.
+#'
+#' Row metadata for microbiome data contains taxonomic annotations at the 
+#' phylum, class, order, family, and genus levels.
```
TuomasBorman: Did I check correctly; rowData was empty?

R/miaTime.R
```diff
@@ -138,3 +138,50 @@ NULL
...
+#'
+#' Quality control metrics include Pearson correlation (>0.98) for HITChip, 
+#' qPCR assay efficiency (>0.99), and technical replicates for HPLC and qPCR.
+#'
+#' Data sources:
+#' - Microbiota HITChip microarray data
+#' - qPCR absolute abundance data
+#' - Chemical profiling data (HPLC, LC-HRMS)
```
TuomasBorman, line 210: metabolite data does not have rownames. Add "feature-1", "feature-2"....

R/miaTime.R
```diff
@@ -138,3 +138,50 @@ NULL
 #' @keywords data
 #'
 NULL
+
+#' @title CrestStudyData
+#' @description
+#' The CrestStudyData includes microbiota and metabolite profiling data 
+#' from 78 Indian individuals (40 males, 38 females) and 78 European individuals
```
TuomasBorman: The dataset seems to have only participants with Indian origin

R/miaTime.R
TuomasBorman: 3. Age groups:
0-10 10--20 20-30 30-40 40-50 50-60 60-70 70-80 80-90

--> Fix 10--20 --> 10-20

R/miaTime.R
TuomasBorman: Check that the description reflects the dataset (not necessarily the study)

TuomasBorman (changes requested):

R/miaTime.R
antagomir: Could be also according to the SurnameYear of the publication (based on first author's name).

R/miaTime.R
antagomir: To check from the Zenodo repo of the publication what is available. If it is not there, we can see if it is possible to add.

R/miaTime.R
antagomir: or feat1, feat2..? (in general "-" is not recommended in names)

R/miaTime.R
0xMuluh: I have checked. Only has the three resolutions oligotype, genus, phylum,

antagomir: Yes but do we have the mappings (rowData)?

R/miaTime.R
0xMuluh: We do have abundance data only on the eu cohort. should this be added in altExp?

R/miaTime.R
0xMuluh: No. i will update the description.

antagomir:
There are unresolved comments. Can you respond, or close if they are clear? Then we can assess merge.

antagomir:
Some checks also failing.

0xMuluh:
> Some checks also failing.

The fail is unrelated. Has to do with gha mac

antagomir (approved):
Seems good!

antagomir:
Ok to merge.

TuomasBorman (changes requested):
Thanks, final things. Sorry for the delay

1. Is there R script available that produces this data? You could add it.

2. TRUE/FALSE values in colData are characters. Convert to logical.

3. Remove extra data. There are additional assays in many experiments. Moreover, there is shannon index in colData. The data could include only the "raw" data to reduce the memory size and also because these results can be calculated from raw data.

3. Convert sample name format from "Sample-1" to "sample_1" (Underscore is usually safer)

4.  `rownames(altExp(tse, 4))` could then be formatted similarly to sample names, i.e., "feature_1" as they are generated by us.

TuomasBorman (approved):
Thanks!

Issue #93, opened 2024-10-08
survival analysis


antagomir:
[Microbiome compositional data analysis for survival studies](https://academic.oup.com/nargab/article/6/2/lqae038/7658053?login=true) - Pujolassos et al. (2024) presents a method for compositional survival analysis for microbiomes.

"The new functions developed for survival analysis have been added to the existing coda4microbiome R package, available at CRAN (https://cran.r-project.org/web/packages/coda4microbiome/index.html). A detailed tutorial about the new functions is available in coda4microbiome blog (https://malucalle.github.io/coda4microbiome/). The data and code for reproducing the analysis are available at DOI 10.5281/zenodo.10552383."

-> Survival analysis functionality should deserve support in miaTime based on the more systematic review.

antagomir:
Let's discuss this.

PR #107, opened 2025-03-20
add survival data

AliHajj20:
I used the data_survival dataset from the following article: https://academic.oup.com/nargab/article/6/2/lqae038/7658053
I converted the dataset to TreeSE and included it in the miaTime package

PR #112, opened 2025-05-06
add getSurvival

0xMuluh:
ping #93


0xMuluh:
This wrapper currently defaults to using pairwise log-ratio (PLR) transformations via `coda4microbiome::coda_coxnet() `, consistent with the approach described in [Calle et al., 2024.](https://doi.org/10.1093/nargab/lqae038)

plr was selected by the authors over clr, ilr, or alr for reasons related to interpretability, sparsity, and model perfomance in high-dimensional microbiome survival analysis. While this default appears justified and performs well empirically, it may not be optimal in all cases.

If plr proves sufficient across expected use cases, we can  add the unit tests around this assumption.

If user needs or new evidence suggest flexibility is needed, we should refactor to allow other transformation via a parameter (e.g., transform.method = "plr", "clr", "ilr", etc.). or `mia::transformAssay()`

Currently the method returns a list of objects including ggplots and dfs/matrices. We can consider `addSurvival()` and store in metadata?

Open to discussion on whether this needs to be prioritized now or flagged for future maintenance..

TuomasBorman (changes requested):
Thanks for this clear PR!

One thing to discuss is how much functionality we want to utilize from coda4microbiome package. If there are plans for broader interlinking, this PR might be fine. However, under the hood, coda_coxnet() is utilizing survival::coxph() and glmnet::cv.glmnet() and survminer for plotting.

The implementation in coda_coxnet() is not complex but includes some overhead that -- in my opinion -- does not necessarily fit well to our current design choices (transformation, printing, plotting).

Instead, this functionality for cox model could be implemented directly in miaTime package. We could take inspiration from the coda4microbiome package, but it seems that there is licence mismatch so this should be taken into account.

antagomir:
Thanks!

Indeed, I would not wrap coda4microbiome. We may like to test and run standard survival models (Cox) also with other transformation. 

So let's keep separate:
- data preparation (e.g. plr -> transformAssay)
- statistical analysis / testing (e.g. Cox model, and glmnet if needed for feature selection)
- summaries (e.g. tables)
- visualizations (-> miaViz?)

This can take inspiration from coda4microbiome and cite it but Cox & GLM are more general and previously published methods. Moreover, the coda4microbiome paper did not include many comparisons to alternative transformations & methods. Our implementation should be modular and support different combinations of choices (e.g. transformation -> testing -> presentation).

The miaTime already includes data from the paper as example data, that could be used to verify the workflow (should yield the same result).

R/getSurvival.R
```diff
@@ -0,0 +1,285 @@
...
+#' 
+#' @param nvar \code{Integer scalar}. Optional. Maximum number of variables 
+#' (log-ratios) to include in the model. (Default: \code{NULL})
+#' 
+#' @param coef_threshold \code{Numeric scalar}. Minimum absolute value for a 
+#' coefficient to be included in the final model. (Default: \code{0})
+#' 
+#' @inheritParams addBaselineDivergence
+#'
+#' @return A list with model summaries:
```
0xMuluh, line 55: can define `addSurvival()`. 
1. Consider adding risk.scores to `colData()`. 
Remaining metrics can be added as attributes to `rowData()/colData()`
or 
2. Just add to `metadata()`

R/getSurvival.R
```diff
@@ -0,0 +1,285 @@
...
+
+############################# internal helpers #################################
+
+.check_data_for_survival <- function(x, time.col, 
+    status.col, covar.cols, assay.type, ...) 
+{
+    x <- .check_and_get_altExp(x, ...)
+    .check_input(time.col, list("character scalar"), colnames(colData(x)))
+    .check_input(status.col, list("character scalar"), colnames(colData(x)))
+    .check_assay_present(assay.type, x)
```
0xMuluh, line 136: Consider adding pairwise logratios `plr` to `transformAssay()`

0xMuluh (commented):
Thoughts. @antagomir @TuomasBorman

R/getSurvival.R
```diff
@@ -0,0 +1,288 @@
...
+#' @export
+#' @importFrom survival Surv coxph
+#' @importFrom glmnet cv.glmnet glmnet
+setMethod("getSurvival", signature(x = "SummarizedExperiment"),
+    function(
+        x,
+        time.col,
+        status.col,
+        assay.type = "counts",
+        covar.cols = NULL,
```
TuomasBorman: Check argument names that we are already using. I think we are using `col.var` which would suit also here

R/getSurvival.R
```diff
@@ -0,0 +1,288 @@
...
+#'   \item \code{alpha}: \code{Numeric scalar}. Elastic net mixing parameter: 
+#'   1 = Lasso, 0 = Ridge. (Default: \code{0.9})
+#'
+#'   \item \code{nfolds}: \code{Integer scalar}. Number of cross-validation 
+#'   folds for \code{cv.glmnet}. (Default: \code{10})
+#' 
+#'   \item \code{nvar}: \code{Integer scalar}. Optional. Maximum number of 
+#'   variables (log-ratios) to include in the model. (Default: \code{NULL})
+#' 
+#'   \item \code{coef_threshold}: \code{Numeric scalar}. Minimum absolute value
```
TuomasBorman: coef.threshold

R/getSurvival.R
```diff
@@ -0,0 +1,288 @@
...
+        covar = covar
+    )
+}
+
+.calc_survival <- function(mat, time, status, covar = NULL,
+    penalized = TRUE, alpha = 0.9, nfolds = 10,
+    lambda = "lambda.1se", nvar = NULL,
+    coef_threshold = 0, ...) 
+{
+    y <- survival::Surv(time, status)
```
TuomasBorman: Use importFrom

R/getSurvival.R
```diff
@@ -0,0 +1,288 @@
...
+        covar <- as.data.frame(colData(x)[, covar.cols, drop = FALSE])
+    }
+    
+    list(
+        mat = mat,
+        time = time,
+        status = status,
+        covar = covar
+    )
+}
```
TuomasBorman: Use the same coding style as elsewhere

R/getSurvival.R
```diff
@@ -0,0 +1,288 @@
...
+#' # tse <- SurvivalData
+#' # getSurvival(tse, time.col = "T1Dweek", status.col = "T1D", 
+#' # covar.cols = c("Sex", "Antibiotics"))
+#'
+NULL
+
+#' @rdname getSurvival
+#' @export
+#' @importFrom survival Surv coxph
+#' @importFrom glmnet cv.glmnet glmnet
```
TuomasBorman: Move these above functions where these are actually used. Helps in updating

R/getSurvival.R
```diff
@@ -0,0 +1,288 @@
...
+    
+    list(
+        mat = mat,
+        time = time,
+        status = status,
+        covar = covar
+    )
+}
+
+.calc_survival <- function(mat, time, status, covar = NULL,
```
TuomasBorman: You can split this function to smaller chunks

R/getSurvival.R
```diff
@@ -0,0 +1,288 @@
...
+        ## apply coef_threshold
+        selected <- which(abs(coefs) > coef_threshold)
+        coef.filtered <- coefs[selected]
+        
+        ## normalize
+        if (length(coef.filtered) > 0) {
+            coef.filtered <- 2 * coef.filtered / sum(abs(coef.filtered))
+        }
+        
+        risk.scores <- predict(fit, type = "lp")
```
TuomasBorman: In variable names, use risk_scores

R/getSurvival.R
TuomasBorman: Could also be defined as a formula

R/getSurvival.R
```diff
@@ -0,0 +1,288 @@
...
+        res <- .calc_survival(
+            mat = dat$mat,
+            time = dat$time,
+            status = dat$status,
+            covar = dat$covar,
+            ...
+        )
+              
+        return(res)
+          }
```
TuomasBorman: Indentation off

TuomasBorman (changes requested):
Thanks! Can you also add unit tests?

R/getSurvival.R
TuomasBorman: addSurvival() would be natural extension. attributes are not that easily accessible so I would avoid that.

metadata would probably be suitable

0xMuluh:
> Thanks! Can you also add unit tests?

unit tests added after #107

R/getSurvival.R
```diff
@@ -0,0 +1,352 @@
...
+#'
+#' @export
+#'
+#' @title
+#' Survival analysis
+#'
+#' @description
+#' Fit a (penalized) Cox proportional hazards model on microbiome data contained
+#' in a SummarizedExperiment object. Data transformations (e.g. pairwise
+#' log-ratios) should be handled upstream
```
TuomasBorman, line 12: It is not evident what this method does. 

Can you add Detail section?

R/getSurvival.R
```diff
@@ -0,0 +1,352 @@
...
+    coefs <- .filter_and_normalize_coefs(coefs, ...)
+    #  Compute risk scores for all samples
+    risk_scores <- predict(fit, type = "lp")
+    #  Compute apparent concordance index
+    c_index <- .compute_c_index(risk_scores, y, fit, penalized = FALSE)
+
+    # Return results as a list
+    res <- list(
+        coefficients = coefs,
+        risk_scores = as.numeric(risk_scores),
```
TuomasBorman: Coefficients coul dbe added to rowData and risk_scores to colData in addSurvival

R/getSurvival.R
```diff
@@ -0,0 +1,352 @@
...
+# Removes coefficients with magnitude below a threshold and rescales the
+# remaining coefficients.
+.filter_and_normalize_coefs <- function(coefs, coef.threshold = 0, ...) {
+    # Identify coefficients exceeding the threshold in absolute value
+    # to remove very small coefficients so that we can focus on meaningful
+    # predictors
+    res <- coefs[ abs(coefs) > coef.threshold ]
+    # Rescale the remaining coefficients
+    if( length(res) > 0L ){
+        res <- 2*res / sum(abs(res))
```
TuomasBorman, line 309: What is the intention of multiplying the values with 2?

TuomasBorman (changes requested):
Thanks!

R/getSurvival.R
antagomir: I think that [mia PR #740](https://github.com/microbiome/mia/pull/740) already provides this.

antagomir:
In manpage description clearly cite the publication where the data is from (note that there is a special way to add citations in roxygen). Now you only refer to R package. But there is also formal publication to cite.

Cite all references in the manpage text so that it is clear why they are added on the reference list.

Running the manpage examples gives error

> The assay contains already only strictly positive values. Pseudocount is not added.
> Error in list(coefficients = coefs, risk_scores = risk_scores, c_index = c_index,  : 
>   argument 6 is empty

Either the function name should be getCox or something explicit, or if it is getSurvival then there should be a method argument that specific what survival method one wants to use.

antagomir:
These are time-to-event models. How about renaming status.col into event.col?

antagomir:
getSurvival has 

 A list with model summaries:

        • ‘coefficients’: estimated model coefficients

        • ‘selected_features’: nonzero features in penalized model

        • ‘risk_scores’: predicted risk scores

        • ‘cindex.apparent’: apparent concordance index

        • ‘cv.cindex.mean’: mean cross-validated C-index (if penalized)

        • ‘cv.cindex.sd’: SD of cross-validated C-index (if penalized)

        • ‘fit’: fitted model object (coxph or cv.glmnet)


-> Harmonize to use "_" as the separator for all (do not mix)

-> cindex.apparent could be C.apparent, or just C ?

-> For clarity I would rename cv.cindex.mean to C.mean.cv and same for the SD ting

-> should selected_features be just nonzero_features if these are the same thing

-> coefficients is usually just coef in many R methods -> harmonize?

R/getSurvival.R
0xMuluh: to have interpretable contrasts that live in space `[-1, 1]`.  after the operation L1 norm is `2`

0xMuluh:
> In manpage description clearly cite the publication where the data is from (note that there is a special way to add citations in roxygen). Now you only refer to R package. But there is also formal publication to cite.

The data has not been used. Will use this data instead: https://github.com/microbiome/miaTime/commit/b752170c2294d82172d5d0d77cc56b3dff00a00d

0xMuluh:
> Running the manpage examples gives error
> 
> > The assay contains already only strictly positive values. Pseudocount is not added.
> > Error in list(coefficients = coefs, risk_scores = risk_scores, c_index = c_index,  :
> > argument 6 is empty

Thanks, there was a typo `,` in the return list.

